PR #21, opened 2022-06-22
Add dev flavour to Android

gewfy:
Does not work, you know why @Pelsin


kajsaunge:
Nice! Running `yarn android` builds on my device 😃

Pelsin:
There is no mocked-google-services that has matching package_name for the dev variant.
We also have not setup any demo project for emulator to connect to, when we tried before the auth was connected to a live project.

Pelsin:
Try now :) @gewfy

PR #103, opened 2022-07-29
Temple / call creation with firestore doc subscription 

Nipher:
- [x] Endpoint to create a DailyCo room and a live-conent-session to be used among participants (with firestore onSnapshot)
- [x] List of rooms
  - [x] Fetch and render rooms
  - [x] Loader/refresher
- [x] Create room
  - [x] Create room and re-fetch rooms list
  - [x] Allow custom name (somehow doesn't work to do it like dailyco api docs says)
  - [x] Create live content doc (firebase complaining about app not initialized)
- [x] Join room
  - [x]  Live content doc snapshot sync
    - [x] Emulator
    - [x] Staging
  - [x] Join room call on multiple devices (test)


Nipher:
Seems like when I run it locally the client firestore is not connecting to the emulator, can't find documents, etc...
not sure why, it worked ok a while ago

Nipher:
Tested running onSnapshot API against staging and it works fine.

Not sure what's up, it seems like the app is not connecting to the Firestore emulator, the firebase rules seem correct, the auth works through the emulator, but firestore somehow "fails" silently, almost like it is connecting to some other firestore with no data in it. 🤔

client/__mocks__/@react-native-firebase/firestore.js
```diff
@@ -0,0 +1,14 @@
+const mockFirestore = {
```
Nipher, line 1: Maybe possible and better to use the real emulator with "seeded mock data" instead 🤓

client/src/routes/Temples/Temples.tsx
```diff
@@ -0,0 +1,91 @@
...
+  padding: SPACINGS.SIXTEEN,
+  borderRadius: SPACINGS.TWELVE,
+  backgroundColor: COLORS.WHITE_TRANSPARENT,
+});
+
+type ScreenNavigationProps = NativeStackNavigationProp<ScreenProps>;
+
+const Temples = () => {
+  const {t} = useTranslation(NS.SCREEN.TEMPLES);
+  const {fetchTemples, isLoading, addTemple, temples} = useTemples();
```
Nipher, line 40: thoughts?

client/android/app/build.gradle
```diff
@@ -282,6 +282,12 @@ android {
 }
 
 dependencies {
+    // Import the BoM for the Firebase platform
+    implementation platform('com.google.firebase:firebase-bom:30.3.0')
+
+    implementation 'com.google.firebase:firebase-auth'
+    implementation 'com.google.firebase:firebase-firestore'
```
gewfy, line 289: I'm pretty sure that these lines are not needed as React Native Firebase is auto linked including the Firebase native lib. Do you reckon why this was added?

Nipher: Probably from one of firebase docs I checked while trying to understand how things work and make it work on my machine

But it was probably a native code documentation, React Native Firebase doesn't mentions it, could probably be removed 👍

gewfy (approved):
e578293034d9d0699efa1bc4c269088fc68b68a9_00 (1)

gewfy (approved):
MYUM

gewfy (approved):
4Bi1

PR #159, opened 2022-08-19
Fix env variables in functions

### LINKED ISSUE #156: Firbease functions doesn't load `.env`
### What brings you here and how are you?

While coding with 29k friends we found something
and I feel okay, glad that we have a way around the issue for now.

### Description

Firbease functions doesn't seem to be loading `.env` like it is said it would on the [docs webpage](https://firebase.google.com/docs/functions/config-env).

For now, we are using `dotenv` as a dependency ourselves and importing it to load them, which should be unnecessary according to the doc linked above.

Expected behavior:

When running `yarn start` in `functions`we expect our `.env` values to be loaded.

Actual behavior:

We don't get the values loaded into `process.env`.

Reproduces how often:

Everytime when running `yarn start`

gewfy:
Adds env validation (and types)

Closes #156


swemail (approved):
Nice!

PR #162, opened 2022-08-19
Docs directory

kajsaunge:
Adds a docs directory for contributions and code of conduct and future docs likes styleguides, architecture maps etc.
Breaks out readme instructions.


gewfy:
I think we should have client and functions in the same document, but under docs and instead split up and linked to from README.md:
* Installation
* Local development
* Build and Deploy
* Style guide
* Translations
* Contributing
* Code of Conduct

gewfy (approved):

PR #163, opened 2022-08-21
Device language detection

gewfy:
Set i18next to use the devices preferred language.

Requires native release


client/src/lib/i18n/index.ts
```diff
@@ -1,19 +1,21 @@
 import 'intl-pluralrules';
 import i18next from 'i18next';
 import {initReactI18next} from 'react-i18next';
+import {findBestAvailableLanguage} from 'react-native-localize';
```
swemail, line 4: Nice function, any reason we didn't use it before?

swemail (approved):

client/src/lib/i18n/index.ts
gewfy: No idea actually!

PR #181, opened 2022-08-22
Removes firebase-tools and node from root

kajsaunge:
To avoid dependabot updates, [like this one](https://github.com/29ki/29k/pull/168)


kajsaunge:
Closing this as it is being used not only for functions, it's also used for CMS etc. [See usage](https://github.com/29ki/29k/blob/main/.github/workflows/deploy-functions.yml#L60)

PR #185, opened 2022-08-22
Also cache Pods folder in GitHub actions

gewfy:


gewfy:
This clearly didn't work and just made the checks a lot slower :P

PR #237, opened 2022-08-31
UI library and components updates

kajsaunge:
- Add icons to ui lib
- Update buttons and cards ui lib
- Add smoother click effect on buttons and cards
- Align color usage on icons
- Minor copy change


client/src/common/components/Cards/Card.tsx
```diff
@@ -71,7 +71,7 @@ export const Card: React.FC<CardProps> = ({
   children,
 }) => (
   <Shadow>
-    <Wrapper onPress={onPress}>
+    <Wrapper onPress={onPress} activeOpacity={0.7}>
```
gewfy, line 74: Nitpick… For the future I suspect that we will miss adding this prop at a lot of places. Probably not a biggie. But maybe good to find a global/reusable solution to this? Like we had in the old app, a reusable component. Or some other way.

gewfy (approved):
LGTM!

client/src/common/components/Cards/Card.tsx
kajsaunge: Yeah! Crossed my mind, thanks for pointing it out! Love nitpicks 😊

PR #245, opened 2022-09-01
Draft: temple content

kajsaunge:


client/src/routes/Temple/hooks/useTemple.ts
```diff
@@ -1,11 +1,12 @@
...
+import {useRecoilState} from 'recoil';
 import {templeAtom} from '../state/state';
 import {Temple} from '../../../../../shared/src/types/Temple';
+import * as templeApi from '../../Temples/api/temple';
 
 const useTemple = () => {
-  const setTempleState = useSetRecoilState(templeAtom);
+  const [temple, setTempleState] = useRecoilState(templeAtom);
 
   const subscribeTemple = useCallback(
```
Nipher, line 11: Currently we always do the `subscribeTemple`  call in the render of a screen (`useEffect`).

Maybe would actually make sense to change the API here to just be `const { navigateToIndex, someOtherMethod } = useTemple('temple-id');`.
)`

Nipher: and move the subscribe/unsubscribe `useEffect` part into the hook itself

gewfy (approved):

PR #262, opened 2022-09-05
Functions REST authorization

gewfy:
Adds a simple koa middleware for authorizing client requests

### DIFF
--- a/functions/src/api/lib/firebaseAuth.ts
+++ b/functions/src/api/lib/firebaseAuth.ts
@@ -1,0 +1,24 @@
+import {getAuth} from 'firebase-admin/auth';
+import {Context, Next} from 'koa';
+
+const firebaseAuth = () => async (ctx: Context, next: Next) => {
+  const {headers} = ctx;
+
+  if (headers?.authorization) {
+    // deny unsupported token format or empty token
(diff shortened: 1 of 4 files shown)


Nipher (approved):

functions/src/api/lib/firebaseAuth.ts
```diff
@@ -0,0 +1,24 @@
+import {getAuth} from 'firebase-admin/auth';
+import {Context, Next} from 'koa';
+
+const firebaseAuth = () => async (ctx: Context, next: Next) => {
+  const {headers} = ctx;
+
+  if (headers?.authorization) {
+    // deny unsupported token format or empty token
+    const [type, token] = headers.authorization.split(' ', 2);
+    if ((type != 'Bearer' && type != 'bearer') || !token) {
```
swemail, line 10: Nit pick, the linter must be borked here, should be `!==`

swemail (commented):
Nice work!

functions/src/api/lib/firebaseAuth.ts
gewfy: True! And this is code from the old app... interesting 🤔

PR #265, opened 2022-09-05
Sets facilitator and show controls accordingly

Nipher:


gewfy:
Nice build on each other here <3 <3 <3

functions/src/api/temples/index.ts
```diff
@@ -38,6 +38,7 @@ templesRouter.post('/', validator({body: CreateTempleData}), async ctx => {
     index: 0,
     playing: false,
     contentId,
+    facilitator: ctx.user.id,
```
gewfy, line 41: I would like some kind of middleware like the Koa middleware, that requires an authorized request here. Basically making sure that `ctx.user` is set

swemail (approved):

PR #270, opened 2022-09-06
Updated typings

swemail:


swemail:
The significant change here is to not use the constants we had before, rather use strings directly since they are type checked. This change is needed for #261

Nipher (approved):
Niiice!

PR #274, opened 2022-09-08
Facilitator slide

gewfy:


functions/src/api/temples/index.spec.ts
```diff
@@ -58,6 +61,14 @@ const mockServer = createMockServer(
...
 );
 
+beforeEach(async () => {
+  await Promise.all(
+    temples.map(temple =>
+      firestore().collection('temples').doc(temple.id).set(temple),
+    ),
+  );
+});
+
```
gewfy, line 71: Like it! Merge all the things!

Nipher:
The call doesn't seem to work anymore on my ios emulator for some reason, neither the disable/enable camera/mic...

Will build to my android now to see what happens

Nipher:
Rebuilt the app now and it works 🎉

Nipher:
@gewfy @swemail how do you feel about [481719c](https://github.com/29ki/29k/pull/274/commits/481719c1160f76e705a3e3d6709d30759e1b6e04)?

My intention there was to provide a way for any component/content to let the selector know who's on spotlight, so we can adapt/format the list of participants with that in mind, instead of doing that on the view layer.

swemail (approved):

PR #297, opened 2022-09-12
Video player with playback time shift

gewfy:


swemail (approved):
So nice, works like a charm

gewfy:
> So nice, works like a charm

Thanks! There are som edge-cases with the time shift that might be worth fixing:
* If the facilitator never starts the video, but a user joins late - the time shift will be based on the last slide navigation
* If the facilitator pause and play a couple of times, and then a user joins late - the time shift will be based last playing and not from the beginning
* If the users local time is drifted everything will be borked.

So not sure if we should have time shift enabled or not.

PR #307, opened 2022-09-14
Portal intro

kajsaunge:


Nipher:
Now that the session only starts if the facilitator presses "start" we could actually merge it I believe.
there's just no way to set the temple state back to "not started", which means it would execute the portal animation straight away.

swemail:
I think it's fine for now. Let's do the reset in another pr if needed.

Nipher (dismissed):

Nipher (dismissed):

Nipher (approved):

PR #320, opened 2022-09-15
Show dark status bar text as default

swemail:
On iOS with dark mode

On Android keeps the contrast on the status bar since the status bar is always showing


kajsaunge (commented):
Looks good! Maybe adding some background, like android. What happens now is this that the statusbar text collides with app content

gewfy:
Sorry I completed this yesterday (should have marked that in notion). This is not the solution as the app still beleives it is in dark mode. See https://github.com/29ki/29k/pull/317

PR #325, opened 2022-09-15
Do not mute when start playing on sharing slide

swemail:


client/src/routes/Temple/hooks/useMuteAudioListener.spec.tsx
```diff
@@ -37,14 +41,18 @@ describe('useMuteAudioListener', () => {
       </RecoilRoot>
     );
 
+    mockUseTempleExercise.mockReturnValue({
+      slide: {current: {type: 'reflection'}},
```
Nipher, line 45: > it('should toggle audio when state is playing and current slide is not reflection',

should  it not play when reflection? the description doesn't seem to match the assertion 🤔

swemail: Right you are, changed it

Nipher (approved):
Nice 🤘

gewfy (approved):
The name of this hook is still a bit vague to me. But let's merge for now!

PR #334, opened 2022-09-15
Ignore silent switch when playing videos

gewfy:


swemail (approved):
Probably why I didn't hear the video, I always have silent switch on 😄

PR #335, opened 2022-09-15
Small tweaks

gewfy:
- [x] [Add some gutters to slide headings](https://github.com/29ki/29k/pull/335/commits/68e2c3f43d51c79dfe27cd735d00730eb854be7b)
- [x] [Text input and button height alignment](https://github.com/29ki/29k/pull/335/commits/28ec7b22bdb8558fefbe8e5caad5eaf2d11a86a0)
- [x] [Add loading state to create temple button](https://github.com/29ki/29k/pull/335/commits/7d5db9e4a13cd2598e143a795685d19781bf4b91)
- [x] [Only show delete on temples owned by the user](https://github.com/29ki/29k/pull/335/commits/b554f4e67c6cfee9486602fbec232d9eb0d9e940)

PR #336, opened 2022-09-16
Button component visual state and props

kajsaunge:
- Add visual que for disabled buttons
- Remove primary prop for now since it's not in the design


gewfy:
Like that we wanted to solve the same issue. We're going to have a conflict around disabled state here :P 
https://github.com/29ki/29k/pull/335/files#diff-75478a124406d233ef8c5409b9adc09161e22a9e436343a91d28b2e05877c4e2R36

PR #339, opened 2022-09-16
Image support in content

swemail:


gewfy (dismissed):
Nice re-use of fields! Did you check if images renders correctly in client? :P

cms/src/fields/slides.js
```diff
@@ -7,13 +7,26 @@ export const SLIDE_TYPES = {
...
 };
 
+const CONTENT_IMAGE_FIELD = {
+  label: 'Image',
+  name: 'image',
+  widget: 'object',
+  collapsed: true,
+  required: false,
+  i18n: false,
+  fields: [{...IMAGE_FIELD, name: 'source', required: false}],
```
gewfy, line 17: Maybe we should change the `IMAGE_FIELD` and `VIDEO_FIELD` to be `{ source: '', description: ''}` as default?

gewfy: And for `VIDEO_FIELD`:
{ source: '', description: '', thumbnail: '' }

swemail:
> Yes, it renders correctly with the images I tested. Looked ok

cms/src/fields/slides.js
swemail: I'll add it now and we can use it later

gewfy:
I want to add a 29k logo on the last slide when this is out 😍

gewfy (approved):

PR #354, opened 2022-09-19
Portal bugs

kajsaunge:
- [x] Fix video full width
- [x] Update participants count

### DIFF
--- a/client/src/lib/navigation/Stacks.tsx
+++ b/client/src/lib/navigation/Stacks.tsx
@@ -28,11 +28,7 @@ const TempleStackWrapper = () => (
       }}>
       <TempleStack.Screen name={'ChangingRoom'} component={ChangingRoom} />
       <TempleStack.Screen name={'Portal'} component={Portal} />
-      <TempleStack.Screen
-        options={{gestureEnabled: false}}
-        name={'Temple'}
-        component={Temple}
-      />
+      <TempleStack.Screen name={'Temple'} component={Temple} />
     </TempleStack.Navigator>
   </DailyProvider>
 );
--- a/client/src/lib/navigation/Stacks.tsx
+++ b/client/src/lib/navigation/Stacks.tsx
@@ -28,7 +28,11 @@ const TempleStackWrapper = () => (
       }}>
       <TempleStack.Screen name={'ChangingRoom'} component={ChangingRoom} />
       <TempleStack.Screen name={'Portal'} component={Portal} />
-      <TempleStack.Screen name={'Temple'} component={Temple} />
+      <TempleStack.Screen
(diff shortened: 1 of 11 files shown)


gewfy:
Add this to the video component to get the audio to be properly mixed between daily and video 
https://github.com/29ki/29k/blob/main/client/src/routes/Temple/components/Slides/Slides/Blocks/Video.tsx#L56

client/src/lib/navigation/Stacks.tsx
```diff
@@ -28,7 +28,11 @@ const TempleStackWrapper = () => (
       }}>
       <TempleStack.Screen name={'ChangingRoom'} component={ChangingRoom} />
       <TempleStack.Screen name={'Portal'} component={Portal} />
-      <TempleStack.Screen name={'Temple'} component={Temple} />
+      <TempleStack.Screen
+        options={{gestureEnabled: false}}
```
gewfy, line 32: Pretty sure that this won't disable the hardware back-button on Android which would lead to the same issue as swiping back. May I suggest that we don't have this as a separate route and instead render the portal on top of call until it's started?

swemail:
> Add this to the video component to get the audio to be properly mixed between daily and video https://github.com/29ki/29k/blob/main/client/src/routes/Temple/components/Slides/Slides/Blocks/Video.tsx#L56

It's merged in another branch AWS

client/src/lib/navigation/Stacks.tsx
kajsaunge: Great point, let's do that instead. Thanks @gewfy 🏓

Nipher: how is that going to fix anything? won't you be able to navigate back from the portal anyways?
maybe I misunderstood what the issue is... 🤔

Nipher: navigate back -> join portal again 🤷 

what are we trying to fix exactly?

client/src/routes/Temple/DailyProvider.tsx
```diff
@@ -18,15 +18,17 @@ import {
   videoSharingAtom,
   activeParticipants,
 } from './state/state';
+import {DailyUserData} from '../../../../shared/src/types/Temple';
 
 export type DailyProviderTypes = {
   call?: DailyCall;
   preJoinMeeting: (url: string) => Promise<void>;
-  joinMeeting: () => Promise<void>;
+  joinMeeting: (userData: DailyUserData) => Promise<void>;
```
gewfy, line 26: Did we forget to remove it here?

gewfy (approved):

PR #404, opened 2022-09-20
Button component refactor

kajsaunge:
- Creates a base button to handle states and styles.
- Updates colors variables

We have a bunch of button states and variations - can we create a base component to handle all variations?
Also the IconButton shares a lot of these styles and states, would be nice to reuse the same base.

This draft pr is one step to get the different states for the button component.
image
image


swemail (approved):
Think it's nice and good!

PR #409, opened 2022-09-21
Wait for user to be authenticated before fetching temples

swemail:


gewfy (commented):
I'm suspecting that this will be a more general problem. So I was thinking that api requests should be handled through a hook. That way we could make it reactive and basically only do api-requests when authenticated.
Waiting for authentication could also be an argument when setting up the hook. Wdys?

gewfy (approved):
Actually, maybe we should ship this now before the user test, as it might mess things up.

PR #420, opened 2022-09-22
Keep button alignment for all slide states

kajsaunge:
Swap prev and next buttons to invisible disabled  button on first and last slides to keep buttons in place. 

**Before**

<br />

**After**


swemail (approved):
I like it, simplest way to know the size

PR #426, opened 2022-09-22
Update typography components

kajsaunge:
- Adds Display component
- Align text and headings according to design
- Update where used


swemail (approved):
Love it! Naming it so we don't have to check how big a `B2` is 👍

PR #468, opened 2022-09-30
Update gutter spacings

kajsaunge:
- Add small and big size to gutters
- Adjust visual indication for media toggles
- Remove layout responsibility from buttons
- Update ui lib to visualize gutters, wrap buttons to not allow stretching and add text layout


gewfy (approved):
Nice! LGTM

PR #491, opened 2022-10-03
Minor UI fixes

kajsaunge:
- Update background on temples route
- Align spacing in portal
-  Hitslop to min 44 on buttons

### DIFF
--- a/client/src/common/components/Buttons/BaseButton.tsx
+++ b/client/src/common/components/Buttons/BaseButton.tsx
@@ -18,23 +18,28 @@ export type BaseButtonProps = {
   active?: boolean;
 };
 
-const BaseButton = styled(TouchableOpacity)<BaseButtonProps>(
-  ({variant, small, elevated, disabled, active}) => ({
-    backgroundColor: disabled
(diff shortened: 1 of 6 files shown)


client/src/common/components/Buttons/BaseButton.tsx
```diff
@@ -18,23 +18,28 @@ export type BaseButtonProps = {
...
-      ? COLORS.WHITE
-      : COLORS.PRIMARY,
-    borderRadius: small ? SPACINGS.TWELVE : SPACINGS.SIXTEEN,
-    alignItems: 'center',
-    justifyContent: 'center',
-    flexDirection: 'row',
-    ...(elevated ? SETTINGS.BOXSHADOW : {}),
-  }),
-);
+const BaseButton = styled(TouchableOpacity).attrs({
```
kajsaunge, line 21: @gewfy do you want to help me understand hitSlop? I've tried testing but can't validate the impact of this change. Also should be specific depending on type of button, If I understand it right (which I probably don't)

swemail (approved):

PR #494, opened 2022-10-03
Adding sound to portal

swemail:
- [x] Enable audio to video in the cms
- [ ] ~~Play audio over portal intro video (Using React Native Video)~~
- [x] Play audio over portal intro video (Using dedicated audio library??)
- [x] Fade audio files
- [x] Use correct audio file


swemail:
Sound is not repeating smoothly on iOS, is it as simple as the format? [issue](https://github.com/react-native-video/react-native-video/issues/1392)

client/src/lib/audio/components/Audio.tsx
```diff
@@ -0,0 +1,39 @@
+import React, {useEffect, useImperativeHandle} from 'react';
+import type RNSound from 'react-native-sound';
+import useSound from '../hooks/useAudio';
```
swemail, line 3: You mean useAudio right 😉

gewfy (approved):

PR #540, opened 2022-10-07
Update counter with formats and optimizations

Nipher:
Screenshot 2022-10-07 at 14 32 03
Screenshot 2022-10-07 at 14 31 15
Screenshot 2022-10-07 at 14 31 08
Screenshot 2022-10-07 at 14 31 03

### DIFF
--- a/client/src/routes/Temple/components/Counter/Counter.tsx
+++ b/client/src/routes/Temple/components/Counter/Counter.tsx
@@ -15,9 +15,20 @@ type CounterProps = {
 const Counter: React.FC<CounterProps> = ({startTime, starting = false}) => {
   const {t} = useTranslation(NS.COMPONENT.COUNTER);
   const [now, setNow] = useState(dayjs());
+  const [active, setActive] = useState(false);
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


client/src/routes/Temple/components/Counter/Counter.tsx
```diff
@@ -17,12 +17,16 @@ const Counter: React.FC<CounterProps> = ({startTime, starting = false}) => {
...
+      return;
+    }
+
     const interval = setInterval(() => {
       setNow(dayjs());
     }, 1000);
 
     return () => clearInterval(interval);
-  }, []);
+  }, [now, startTime]);
```
swemail, line 29: Will this not recreate the intervall for each now? `setNow` will cause it to run again and so it goes

Nipher: true! let me have a look

Nipher: should be fixed now, nice catch

swemail (approved):

PR #591, opened 2022-10-12
Add script to clean daily rooms

Nipher:
### Description of the Change

Deletes all daily rooms when script runs


kajsaunge (approved):
Superrrr! 🙏

PR #606, opened 2022-10-13
Update sessions query

Nipher:
Issue: https://29k.slack.com/archives/C7DTMLM6Y/p1665651504955139

### What brings you here and how are you?

I'm ok, my local dev environment doesn't seem to work, but I'm adding these changes here if anyone can review until we fix it.

### Description of the Change

Adding back functionality from #592


gewfy (approved):
3rd times the charm!

PR #649, opened 2022-10-17
Upgrade of cms type generator

swemail:
Using the fine updates in the netlify-ts made by @gewfy

### DIFF
--- a/cms/generateTypes.ts
+++ b/cms/generateTypes.ts
@@ -1,12 +1,10 @@
 import * as fs from 'fs';
 import * as process from 'child_process';
-import {createNetlifyTypes} from 'netlify-ts';
+import createNetlifyTypes from 'netlify-ts';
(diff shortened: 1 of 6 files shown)


cms/generateTypes.ts
```diff
@@ -1,13 +1,11 @@
...
 import {Collection} from 'netlify-ts/lib/types';
 import {exercises, contributors, files} from './src/collections/collections';
 
 const OUTPUT_PATH = '../shared/src/types/generated';
 
-const clean = (input: string) => input.replaceAll(/_|-|\./g, '');
-
 const createTypeFile = (collection: Collection) => {
   const fileName = collection.label_singular
     ? collection.label_singular.replace(/^[a-z]/g, v => v.toUpperCase())
```
gewfy, line 11: I don't think we need this either :)

gewfy: Oh sorry... never mind :D

PR #654, opened 2022-10-17
Prevent race condition in user re-creation

gewfy:
This prevents users to be re-created when multiple api calls calls for user re-creation in parallel.

### DIFF
--- a/client/src/lib/apiClient/apiClient.spec.ts
+++ b/client/src/lib/apiClient/apiClient.spec.ts
@@ -138,6 +138,18 @@ describe('apiClient', () => {
     });
   });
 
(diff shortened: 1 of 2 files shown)


swemail:
I think this is fine for now, it's like a mutex or lock or something 👍 No problem in a single threaded environment. I think we can rethink how we recreate the user, it's a bit weird now that we create a new user as soon as a call fails. Maybe we could redirect and recreate the user in a more procedural way so calls can fail even if they are made simultaneously without recreating the user.

swemail (approved):

gewfy:
@swemail Agree! Let's revisit this later :)

PR #657, opened 2022-10-18
Promote user to public host

swemail:
### IMPORTANT do not merge this until #698 have been merged
Flow to upgrade user to `publicHost` role


gewfy (changes requested):
There seems to be some `requests` files in functions still. Is that correct?

swemail:
> There seems to be some `requests` files in functions still. Is that correct?

Got lost in translation there, removed now

PR #666, opened 2022-10-18
Session type selection + Session controller/model

Nipher:
Issue: https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#8f849201312a4784bf669096e250b810

### Description of the Change

- Introduces a `type` filed into `sessions`, and a step on session creation to choose who you'd be doing the session with. (aka session types)
- Moves sessions firestore operations to a model layer

### Verification Process

Go through the create session flow and verify that only public session types show up for now and that your session got the `type` field  on the database with the correct value.

### DIFF
--- a/content/src/ui/Component.CreateSessionModal.json
+++ b/content/src/ui/Component.CreateSessionModal.json
@@ -3,6 +3,21 @@
     "selectContent": {
       "title": "Pick an exercise"
     },
+    "selectType": {
+      "title": "Create session with...",
+      "public": {
+        "title": "Anyone",
+        "icon": "https://res.cloudinary.com/twentyninek/image/upload/v1664528333/temp/pure-simple-love_rd2wvb.png"
+      },
+      "private": {
+        "title": "My friends",
+        "icon": "https://res.cloudinary.com/twentyninek/image/upload/v1664528333/temp/pure-simple-love_rd2wvb.png"
+      },
+      "async": {
+        "title": "My self",
+        "icon": "https://res.cloudinary.com/twentyninek/image/upload/v1664528333/temp/pure-simple-love_rd2wvb.png"
+      }
+    },
     "setDateTime": {
       "title": "Session time set up",
       "cta": "Publish"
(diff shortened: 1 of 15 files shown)


gewfy (changes requested):
Nice! Could we already now make sure to filter `/api/sessions` to only show public sharings?

Nipher:
@gewfy you mean [this](https://github.com/29ki/29k/pull/666/files#diff-c15681cbe272ddd42e4a33b584f4bda214f0130b905c5ed30f28a51fb8a4cf04R59)? 👍 

I'll also follow up, adding specs and implementing what we talked about on [this thread](https://29k.slack.com/archives/C0409B7LGPL/p1666090455781549)

Nipher:
and yes, firestore indexes

functions/src/models/session.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+      firestore().collection('sessions').doc(session.id).set(session),
+    ),
+  );
+});
+
+afterEach(() => {
+  jest.clearAllMocks();
+});
+
+describe('session model', () => {
```
Nipher, line 79: @swemail I ended up transforming the old route specs into model specs, would love to get your input at some point, maybe there are better ways to test it (or not test it at all 🤷)

swemail: Great! I think this is the only way test queries against the mock, great that you moved it over so we can have it as reference.

Nipher: I see that we've controllers now already, so I think i will move some of the logic into them, so the tests for models would be very straight-forward

swemail: 🥚 sactly, the models will be dumb, but it's not wrong to verify the queries so you kan keep that part of the test I think

Nipher:
Missing to add the firestore indexes, will try to get them after lunch

swemail (approved):
Nice work!

content/src/ui/Component.CreateSessionModal.json
```diff
@@ -3,6 +3,21 @@
...
+      "title": "Create session with...",
+      "public": {
+        "title": "Anyone",
+        "icon": "https://res.cloudinary.com/twentyninek/image/upload/v1664528333/temp/pure-simple-love_rd2wvb.png"
+      },
+      "private": {
+        "title": "My friends",
+        "icon": "https://res.cloudinary.com/twentyninek/image/upload/v1664528333/temp/pure-simple-love_rd2wvb.png"
+      },
+      "async": {
```
gewfy, line 16: I don't think this type will ever happen. It will be more async content than async sessions.

gewfy (approved):
Approved, but would be nice to clean some of the comments up in this or another PR :)

PR #671, opened 2022-10-18
Screen component

kajsaunge:
Common component to keep ui consistent 😃 Simple as possible for current needs


gewfy (changes requested):
Hmm I'm not sure about this one. A couple of things that doesn't make this very reusable
* It assumes that the `<BackButton/>` is within `<Gutters>`. Hence the minus margin I guess?
* Some screen we want the back button to be `position: absolute` so it doesn't affect the content. E.g. `ChangingRoom`.
* `<Screen>` doesn't use this one. When do we specify back in screen and when do use this one? Should they be one?

client/src/routes/Session/IntroPortal.tsx
```diff
@@ -192,12 +188,7 @@ const IntroPortal: React.FC = () => {
         {isFocused && (
           <Content>
             <TopBar>
-              <BackButton
-                noBackground
-                onPress={leaveSessionWithConfirm}
-                Icon={ArrowLeftIcon}
-                fill={themedText}
-              />
+              <BackButton onPress={leaveSessionWithConfirm} fill={themedText} />
```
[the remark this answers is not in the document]
kajsaunge: Theme and theme text color is optional in the CMS. What do you mean themedText is a boolean? 😊

client/src/common/components/Screen/Screen.library.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import {ScrollView} from 'react-native-gesture-handler';
```
gewfy, line 2: `ScrollView` is imported from the wrong place here

client/src/common/components/Screen/Screen.library.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const DefaultWithBackgroundColor = () => (
+  <Screen backgroundColor="red">
+    <Display36>
+      WithBackgroundColor. Lorem Ipsum is simply dummy text of the printing and
+      typesetting industry. Lorem Ipsum has been the industry's standard dummy
+      text ever since the 1500s.
+    </Display36>
+  </Screen>
+);
+export const DefaultWithTextColor = () => (
```
gewfy, line 26: This one doesn't really make sense as there is no visual change in the UI library. How about having a backbutton here too?

client/src/common/components/Screen/Screen.tsx
```diff
@@ -13,64 +17,73 @@ const Wrapper = styled.View<{backgroundColor?: string}>(
...
+    height: safeArea.top + SPACINGS.FOURTYFOUR,
+    width: '100%',
+    justifyContent: 'space-between',
+    position: floating ? 'absolute' : 'relative',
+  }),
 );
 
+type OverlayProps = LinearGradientProps & {safeArea: EdgeInsets};
+
+const Overlay = styled(LinearGradient)<OverlayProps>(({safeArea}) => ({
```
gewfy, line 40: Nice addition :)

client/src/common/components/Screen/Screen.tsx
```diff
@@ -13,64 +17,73 @@ const Wrapper = styled.View<{backgroundColor?: string}>(
...
-      {!noTopBar && (
-        <TopBar
-          onPressBack={onPressBack}
-          hasDarkBackground={hasDarkBackground}
-        />
+      {onPressBack && (
+        <TopBarWrapper floating={floatingTopBar} safeArea={safeArea}>
+          {topBarOverlay && (
+            <Overlay
+              colors={[COLORS.WHITE, COLORS.WHITE_TRANSPARENT]}
```
gewfy, line 73: Should'nt this be based on the `backgroundColor` instead?

client/src/routes/Sessions/Sessions.tsx
```diff
@@ -95,7 +95,7 @@ const Sessions = () => {
   );
 
   return (
-    <Screen noTopBar backgroundColor={COLORS.PURE_WHITE}>
+    <Screen backgroundColor={COLORS.PURE_WHITE}>
```
gewfy, line 98: Shouldn't this be the default color in `Screen` or the default of the native app? Else we need to specify this everywhere? Right?

gewfy (changes requested):
I believe you should add this to all routes so we get into the flow of using it. E.g. the `IntroPortal` (but with the custom back button still), `OutroPortal`, `Session`. Basically every screen :)

client/src/common/components/Screen/Screen.library.tsx
kajsaunge: Hahahaha 🤦‍♀️

client/src/common/components/Screen/Screen.tsx
kajsaunge: I tried but couldn't get it to work - wanna help?

client/src/routes/Sessions/Sessions.tsx
kajsaunge: Only session should have `PURE_WHITE`, all other routes should have `WHITE`, which is defined as default in `Screen`

client/src/common/components/Screen/Screen.tsx
kajsaunge: Ooooh I know now why it faild - I used 'red instead of hex. I'll fix that and push - great that you pointed it uot!

client/src/routes/Sessions/Sessions.tsx
gewfy: Ah ofc! Nice :)

PR #701, opened 2022-10-19
Session deep links

gewfy:
- [x] Dynamic links util
- [x] Create link on session creation
- [x] Update docs about `.env` changes
- [x] Update GitHub secrets
- [x] Update workflows with new `.env` variables


functions/src/api/sessions/index.ts
```diff
@@ -84,12 +85,15 @@ sessionsRouter.post('/', validator({body: CreateSessionSchema}), async ctx => {
     timestamp: Timestamp.now(),
   };
 
+  const link = await createDynamicLink(`sessions/${data.id}`);
```
Nipher, line 88: I imagine that would be the invitecode instead of the id for private sessions in the future? 🤔

Nipher (approved):
super cool!

functions/src/api/sessions/index.ts
Nipher: actually both would work since we could fetch the data from either 🙌

gewfy: Yeah I was thinking the same. That for private sessions we could have another pattern. E.g. `sessions/:sessionId/invitation/:invitationCode` or similar.

gewfy: This is where you later setup the routing in the client:

https://github.com/29ki/29k/blob/08892f96c351a0e71dc1854c441d2b1c65f4b977/client/src/lib/navigation/linking.ts#L10-L19

PR #716, opened 2022-10-20
Restrict access to create session

swemail:
With this the backend won't allow creating sessions without being a `publicHost`


swemail:
> Wouldn't we use that for creating private sessions? which everyone will be able to do? 🤔

You're right, it was not there when I started this 😄 Let's reiterate this. Maybe we need different enpoints?

functions/src/api/lib/restrictAccessToRole.ts
```diff
@@ -0,0 +1,18 @@
+import {Next} from 'koa';
+import {getAuth} from 'firebase-admin/auth';
+import {ROLES} from '../../../../shared/src/types/User';
+import {FirebaseAuthContext} from './firebaseAuth';
+
+const restrictAccessToRole =
+  (role: keyof typeof ROLES) =>
+  async (ctx: FirebaseAuthContext, next: Next) => {
+    const customClaims = (await getAuth().getUser(ctx.user.id)).customClaims;
```
gewfy, line 9: Can't we add the claims on the user context with the existing middleware instead? And keep this one to just check that object?

functions/src/api/sessions/index.ts
```diff
@@ -36,18 +37,23 @@ const CreateSessionSchema = yup.object().shape({
...
-    contentId,
-    type,
-    startTime,
-    language,
-  });
-});
+sessionsRouter.post(
+  '/',
+  validator({body: CreateSessionSchema}),
+  restrictAccessToRole('publicHost'),
```
gewfy, line 43: Really like this way of validating 🙌

gewfy (changes requested):

functions/src/api/lib/restrictAccessToRole.ts
swemail: Super improvement, thanks!

Nipher:
or maybe it is possible to check the session type that was sent through the endpoint and check the necessary user role accordingly?

Either that or two endpoints sounds good to me, no strong opinions

Nipher (approved):

gewfy (approved):
Really nice! 🙌

PR #721, opened 2022-10-24
Sessions actions

kajsaunge:
Issue: [Add join to modal](https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#5efc6cd39a31472aa5e3491421ca091d) + [Open session details modal](https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#04c7e529bec743b9bc0994aa41d03590)

### Description of the Change

- Remove plus icon on join button
- Allow card click to always open session details modal (instead of add to calendar)
- Add join button to session details modal 10 min before session starts


client/src/routes/Sessions/components/SessionModal.tsx
```diff
@@ -58,11 +58,25 @@ const SessionModal = () => {
     useSessionNotificationReminder(session);
 
   const startTime = dayjs(session.startTime);
-
+  const startingNow = dayjs().isAfter(startTime.subtract(10, 'minutes'));
```
swemail, line 61: I would make this into a function, dayjs() is like Date.now but now will be when the component renders not when the button is actually pressed. We only need to know now when the button is actually pressed so no need to do a dyajs() call until needed. So, do `cont startingNow = () => dayjs()...` and then `onPress` do `startingNow() ? ...`

swemail (approved):

swemail (approved):

PR #751, opened 2022-10-25
Markdown code styles

kajsaunge:
Issue: There is no way to distinguish what hosts should SAY and DO. [Task in notion](https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#2852e3bc765b4bbbb3a8b4a8adffb956)

### Description of the Change

This change makes use of the markdown `code_inline` to allow content creators to highlight information that should be done rather than said.

#### Screenshots

| Before | After |
|-|-|
| image | image |

### DIFF
--- a/client/src/common/components/Icons/Info/Info.tsx
+++ b/client/src/common/components/Icons/Info/Info.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import {Path} from 'react-native-svg';
+import {IconType} from '..';
+import {COLORS} from '../../../../../../shared/src/constants/colors';
+import Icon from '../Icon';
+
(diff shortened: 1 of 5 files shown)


kajsaunge:
For some reason the text does not show when making use of the code_inline. Any ideas of why this might be?
image

gewfy:
This is the default rule for `code_inline` in the react-native-markdown-display lib - I beleive it gives a hint about what you need to change in your implementation to get it working :) 
https://github.com/iamacup/react-native-markdown-display/blob/0b0dc3ece52ad9d3db4c9c434883e47baf25ba75/src/lib/renderRules.js#L174-L178

gewfy (approved):
👍

PR #760, opened 2022-10-25
Fix keyboard avoiding in half modals

gewfy:
| Before | After |
|-|-|
| image | Screenshot 2022-10-25 at 17 18 43 |


swemail (approved):
Classic, great that you fixed it!

PR #764, opened 2022-10-26
UI lib - modal and card components

kajsaunge:
Add modals to ui library
Simplify cards library 
Replicate Session card 

| image | image


gewfy (approved):
Nice! :)

PR #769, opened 2022-10-27
Remove need to reauthenticate when upgrading account

swemail:
Explicitly refresh claims after upgrade, then re authentication is not needed


gewfy (approved):
Like it! Nice :)

PR #784, opened 2022-10-27
Generate public host verification links

gewfy:
I haven't tested this locally. But tests and types are saying it should work :D
image


functions/src/lib/slack.spec.ts
```diff
@@ -107,7 +108,7 @@ describe('slack', () => {
             type: 'section',
             text: {
               type: 'mrkdwn',
-              text: '*Accepted <mailto:some@email.com?body=123456|some@email.com> as public host, please send code `123456` to the user.*',
+              text: '*Accepted <mailto:some@email.com?body=123456|some@email.com> as public host, please send code `123456` http://some.deep/verification/link to the user.*',
```
swemail, line 111: Maybe add it also t the mailto body?

swemail: Sorry, my bad, just a test

swemail (approved):
Nice addition this!

functions/src/lib/slack.spec.ts
gewfy: Ah! Will fix :)

PR #809, opened 2022-10-28
Draft: Notify users entering and exiting session

kajsaunge:
Issue: There is no way of knowing when a user leaves and enters the session, they just silently end up as last participant. [notion](https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#915c666ef6d94975ac31cee928e5611d)

### Description of the Change
Adding a short message that a user has joined. Remove within a few secs

#### Current Screenshots, will be iterated 

| iOS | Android |
|-|-|
| image |  |

### DIFF
--- a/client/src/routes/Session/components/Participants/Participants.tsx
+++ b/client/src/routes/Session/components/Participants/Participants.tsx
@@ -1,10 +1,16 @@
-import React, {useState} from 'react';
+import React, {useEffect, useState} from 'react';
 import {FlatList, ListRenderItemInfo} from 'react-native';
 import {DailyParticipant} from '@daily-co/react-native-daily-js';
 import styled from 'styled-components/native';
 import {curry} from 'ramda';
+import Animated, {FadeIn, FadeOut} from 'react-native-reanimated';
+import {useTranslation} from 'react-i18next';
 
 import Participant from './Participant';
+import {Body14} from '../../../../common/components/Typography/Body/Body';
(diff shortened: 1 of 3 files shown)


client/src/routes/Session/components/Participants/Participants.tsx
```diff
@@ -31,6 +55,32 @@ const Participants: React.FC<ParticipantsProps> = ({participants}) => {
...
+      }, 4000);
+      return clearTimeout(joinMessage);
+    }
+    if (participants.length < numberOfParticipants) {
+      setNumberOfParticipants(participants.length);
+      setJoinMessage(t('participant.left'));
+      setTimeout(() => {
+        setJoinMessage('');
+      }, 4000);
+      return clearTimeout(joinMessage);
```
Greenheart, line 80: First of all, this will be a nice UX improvement - thank you! 💚 

Just noticed here that the `clearTimeout()` is trying to clear `joinMessage` which is the string state, rather than references to the active timeouts.

What you probably meant to do is to keep the timeout ids returned by `setTimeout()` on line 77 and line 69 and pass those to `clearTimeout()` instead, right?

Also the cleanup function of the useEffect at line 80 should probably be a callback function, like `() => { clearTimeout() }`. Otherwise it would try to use the result of calling `clearTimeout()` right away when the effect is first created.

kajsaunge:
Closing this one as we decided to use daily context

PR #812, opened 2022-10-31
Deep link env example

kajsaunge:
Issue: missing vars in env.example 

### Description of the Change
Add vars to example


gewfy (approved):
Nice catch!

PR #846, opened 2022-10-31
Rename facilitator to host

Nipher:
Issue: 

#### Screenshots

| iOS | Android |
|-|-|
|  |  |


gewfy (changes requested):
Really nice that you took it all the way! 🙌 Could we call it `hostId` in the sessions docs? To align it with `contentId`, `userIds`, `dailySpotlightId` and so on?

swemail (dismissed):
Nice, have been thinking of this a while now without doing anything about it :)

gewfy (approved):
👌 👌 👌

PR #871, opened 2022-11-03
Session unavailable flow

Nipher:
Issue: [Task in notion ⛑️ ](https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#0e1fa739aff840be9544aaa76ea4344e)

#### Screenshots

| Session not available | Session not found |
| ------- | ------- |
|  | Screenshot 2022-11-08 at 10 57 34 |

### Verification Process

Go through the "add session flow", and join a session typing any non-existent or old session invite code.


content/src/ui/Component.SessionUnavailable.json
```diff
@@ -0,0 +1,11 @@
+{
+  "en": {
+    "description": "The session you're looking for is not available anymore."
+  },
+  "sv": {
+    "description": "The session you're looking for is not available anymore."
+  },
+  "pt": {
+    "description": "The session you're looking for is not available anymore."
```
gewfy, line 9: It's better to not duplicate keys in `sv` and `pt`. Then it's easier to see in the CMS what needs translations as they will be empty.

client/src/routes/Sessions/components/SessionUnavailable.tsx
```diff
@@ -0,0 +1,24 @@
...
+      <Gutters>
+        <Spacer24 />
+        <Body16>{t('description')}</Body16>
+        <Spacer24 />
+      </Gutters>
+    </HalfModal>
+  );
+};
+
+export default SessionUnavailable;
```
gewfy, line 24: Should this be named `SessionUnavailableModal` to conform with the other ones maybe?

gewfy (commented):
I believe there are two cases that this solution doesn't cover:
* If your UI state is stale, and you click the join button 30 min after the session have started or  after the session have ended
* If the session gets deleted after you have joined

How about not solving this in the JoinSessionModal, but instead in `useSubscribeToSession`? Then we could handle all three cases at a single point.

client/src/routes/Sessions/components/JoinSessionModal.tsx
```diff
@@ -33,10 +33,15 @@ const JoinSessionModal = () => {
...
+              const session = await joinSession(value);
+              fetchSessions();
+              goBack();
+              navigate('SessionModal', {session: session});
+            } catch (err) {
+              goBack();
+              navigate('SessionUnavailable');
+            }
           }}
         />
```
gewfy, line 46: Outside of this PR, this modal needs better error handling. There's a [todo about that in notion](https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#843cd87bd79248fe91898095d737f0c8). You can also have a look at how it's solved in host invites https://github.com/29ki/29k/blob/main/client/src/routes/Profile/UpgradeAccount.tsx#L147-L162

Nipher: Not sure if I follow, this is that flow, the difference I see is  that it navigates to a new modal that is specific for that user-flow, rather than render something different in the same modal, is that what you meant by better error handling? 🤔

client/src/routes/Sessions/components/SessionUnavailable.tsx
Nipher: I can rename, should we rename UpgradeAccount as well?

content/src/ui/Component.SessionUnavailable.json
Nipher: Ah, gotcha, makes sense 👍

client/src/routes/Sessions/components/SessionUnavailable.tsx
gewfy: Ah yes, please! Thanks :)

client/src/routes/Sessions/components/JoinSessionModal.tsx
gewfy: So if you write the wrong code manually you will not be able to edit it, you will be thrown to this modal instead? Do I understand this correctly? 
In that case it feels like two different cases. One is dealing with bad user input or general errors in `JoinSessionModal`. The other one is dealing with a session that is old/ended/deleted. 
Did you see my comment here? https://github.com/29ki/29k/pull/871#pullrequestreview-1168444119

Nipher: Ah didn't see that comment, thanks!

The two other cases you mentioned seems super rare to me but maybe it would happen more often than I imagine.

Good point, I just don't see how handling it on `useSubscribeToSession` would help on the flow that is being handled atm.

If you only have a invite code or clicked a link, we don't really use `useSubscribeToSession` right? unless we subscribe to session by invite code or something....

And the endpoint to join the session just fails, it doesn't say why (because no session was found with that token or because session is not available any longer)

We could return different error codes to handle them individually... Can sync with Lisa again about it since we discussed it these days.

Nipher: but handling it on `useSubscribeToSession` would definitely work for the two cases you mentioned 👌

gewfy: Great! So in this PR let's forget about the `useSubscribeToSession` error.

Could you adjust this so instead of being a separate modal, the error is shown where you input the code. And then we deal with old/ended/deleted sessions outside of this one? :)

Nipher: I guess I need to sync with Lisa about it, we agreed on showing a modal just like the SessionModal communicating that the session is not available.

And we still have the two cases, "misstype/invalid invite code" (session not found) and session not available anymore

Nipher: Would the error you mentioned still show the input to type the code again?

gewfy: Sounds good :) I think we could say that session not found and not available is the same thing. But I leave that up to you and Lisa.

gewfy: And if you implement two different errors, please align with how host requests are handled in `UpgradeAccount` :) 
https://github.com/29ki/29k/blob/main/client/src/routes/Profile/UpgradeAccount.tsx#L70-L95

gewfy (approved):
Nice!

PR #872, opened 2022-11-03
Leave daily room before outro portal

Nipher:
Issue:  🗒️  [Notion task](https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#ec87d786480e42d5a9d71a14ac8af53d) 

### Description of the Change

We leave the call before navigating to the outro portal.

### Verification Process

Go through the call, you can join as a participant on the web through the url in the session document, when the app user/participant transitions to the outro, they should drop from the call and not be heard or seen

### DIFF
--- a/client/src/routes/Session/Session.tsx
+++ b/client/src/routes/Session/Session.tsx
@@ -87,8 +87,13 @@ const StyledButton = styled(Button)({
 });
 
 const Session = () => {
-  const {setUserData, toggleAudio, toggleVideo, setSubscribeToAllTracks} =
-    useContext(DailyContext);
+  const {
+    setUserData,
+    toggleAudio,
+    toggleVideo,
+    setSubscribeToAllTracks,
+    leaveMeeting,
+  } = useContext(DailyContext);
   const {
     params: {sessionId},
   } = useRoute<RouteProp<SessionStackProps, 'Session'>>();
@@ -111,9 +116,10 @@ const Session = () => {
 
   useEffect(() => {
     if (session?.ended) {
+      leaveMeeting();
       navigate('OutroPortal');
     }
-  }, [session?.ended, navigate]);
+  }, [session?.ended, navigate, leaveMeeting]);
 
   useEffect(() => {
     setUserData({inPortal: false} as DailyUserData);


client/src/routes/Session/hooks/useLeaveSession.ts
```diff
@@ -1,28 +1,21 @@
...
   const {t} = useTranslation('Component.ConfirmExitSession');
   const {leaveMeeting} = useContext(DailyContext);
-  const {navigate} = useNavigation<ScreenNavigationProps>();
 
   const resetSession = useResetRecoilState(sessionAtom);
 
   const leaveSession = useCallback(async () => {
     await leaveMeeting();
     resetSession();
-    navigate('Sessions');
```
gewfy, line 24: This breaks exiting the session. It needs to navigate here since we use this hook for back buttons/behaviour.

client/src/routes/Session/Session.tsx
```diff
@@ -122,9 +123,10 @@ const Session = () => {
 
   useEffect(() => {
     if (session?.ended) {
-      navigate('OutroPortal');
+      leaveSession();
+      navigate('OutroPortal', {portal: exercise?.outroPortal});
```
gewfy, line 127: Can't we leave this as it was, but just add a `leaveMeeting`?

gewfy (changes requested):

client/src/routes/Session/Session.tsx
Nipher: if you leave the meeting you don't have a session state anymore, and `OutroPortal` needs that data to be able to access the `outroPortal` from the session's exercise.

That's why I decided to pass in the Portal in the route params

gewfy: Yeah exactly, but if we don't use `leaveSession` here and only use `leaveMeeting` we will keep the session state while being in the outro portal too :)

Nipher: and then outro portal would call `leaveSession`, which would call `leaveMeeting`again, even though the user already left? 🤔

Nipher: or just call `resetSession` manually I guess?

Nipher: would be nice to always use the leaveSession hook in the same way when leaving 🤔

gewfy: True true... But I think you can run `leaveMeeting` multiple times, even though it's already left. Not sure what the best solution would be here tbh.

swemail (approved):

gewfy (approved):
Nice! :)

PR #919, opened 2022-11-08
Add storage bucket to local gservices

Nipher:
Easier to copy&paste to use the app locally (dev mode)


gewfy (approved):
Nice catch, I'll add it to the release :)

PR #926, opened 2022-11-08
User name and image in session

kajsaunge:
Issue: [Use name and profile in changing room, in session and in session notises](https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#b45f18466ad64f898be27c61a0ba5ea1)

### Description of the Change

If we have userData
- we pre-fill name field and use the image for the no-camera state in the changing room
- the user image is used image in the session if participant has camera off
- we also use the image in session notifications when user joins or leaves

#### Screenshots of different scenarios

**Changing room**
| iOS | Android |
|-|-|
| image |  image |

**Session**
| iOS | Android |
|-|-|
| image |  image |


gewfy (approved):
Boooootiful!!

PR #939, opened 2022-11-09
Session state label and date fixes

gewfy:
Issue: https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#ffd7292c839b44e0b73539f7f6a3e952

- [x] "has started" label in sessions list
- [x] Date counter showing incorrect countdown when date is next day

| Sessions | Not started | Started |
|-|-|-|
| image | image | image |


client/src/routes/Session/IntroPortal.tsx
```diff
@@ -232,16 +232,17 @@ const IntroPortal: React.FC = () => {
...
                 <Badge themeColor={textColor}>
                   <StatusText themeColor={textColor}>
-                    <Counter
-                      startTime={dayjs(session?.startTime.toDate())}
-                      starting={session?.started}
-                    />
+                    {session?.started ? (
+                      t('counterLabel.started')
+                    ) : (
+                      <Counter startTime={dayjs(session?.startTime.toDate())} />
```
swemail, line 244: This is not new but maybe we could find a way to avoid `dayjs()` calls like this. It consumes memory to do this on every re render. It is not a problem here but could be in other places

swemail (approved):

PR #950, opened 2022-11-09
Move all modals to routes folder

gewfy:
Also moved `ProfileInfo` and `VerificationCode` to `common/components`


swemail (approved):
Much better!

PR #959, opened 2022-11-10
UI LIB: session notification

kajsaunge:
Issue: missing component in ui library

### Description of the Change

- Separates ui and logic
- Add prop for timing and visibility to increase flexibility
- Update ui lib

#### Screenshots 
(Note: notifications are not stretched as in these images)

| iOS | Android |
|-|-|
| image | image |
| iOS after 4 sec | Android after 4 sec |
| image | image |
| iOS with visible prop | Android with visible prop |
| image | image |


client/src/routes/Session/components/SessionNotifications.tsx
```diff
@@ -60,7 +61,7 @@ const SessionNotification: React.FC<{
     };
   }, []);
 
-  if (visible) {
+  if (visible || uilib) {
```
kajsaunge, line 64: Is this ok / horrible / stupid / itchy / great or anything else? 😃

gewfy: Yes this approach is a bit to quirky, and shows the need for a more "stupid" component without the logic. Basically I would separate the rendering logic from the behaviour logic. And then only visualise the component without this logic.

kajsaunge: Makes sense! Thanks @gewfy for taking the time to explain

kajsaunge: @gewfy does the separation makes sense, anything that can be improved?

gewfy: Hmm I still think we should have a component that doesn't include the timeout/useEffect/logic at all. And use that in the UI lib. It doesn't make sense to show different timings in the UI Lib (to me).

kajsaunge: Agree it doesn't make sense to have timed components in the uiLib. Although it'd be nice to show how it animates, since that is still related to ui. But not necessary ofc.

Regarding the component it self, I can't help to think it's kinda nice to have a default timeout that you can alter depending on the use case. Am I out riding a bike here? 🚴‍♀️ .. backwards 😂

PR #960, opened 2022-11-10
Sheet and card modals

gewfy:
# Card sheets
| Add Session | Join Session | Session not found |
|-|-|-|
| image | image | Screenshot 2022-11-17 at 17 13 38 |

# Bottom sheets
| Session Details | Create Session Expanded | Request Host | 
|-|-|-|
| image | image | image |


client/src/lib/navigation/constants/routes.ts
```diff
@@ -13,17 +13,18 @@ export type SessionStackProps = {
   OutroPortal: undefined;
 };
 
+export type AppStackProps = {
+  KillSwitch: undefined;
+  Tabs: NavigatorScreenParams<TabNavigatorProps>;
+  SessionStack: NavigatorScreenParams<SessionStackProps>;
+};
+
 export type ModalStackProps = {
```
swemail, line 22: Should `ModalStackProps` be the name of our main Stack? It's like `App` is a modal. Isn't `RootStackProps` for this better?

swemail (approved):
Work of art, as always!

client/src/lib/navigation/constants/routes.ts
gewfy: I pondered the same. But it doesn't make sense when using these types to navigate later. I mean you navigate in the modal stack, not the root stack? In this case App is actually not a modal. The first child of the Modal Stack is what's being rendered behind the modals basically. Ehm, did that make sense? :D

swemail: It makes sort of sense 😂 Let's have it as this and feel it in action

PR #961, opened 2022-11-10
Distinguish between fully supported and less supported languages

gewfy:
- [x] Make English and Portuguese fully supported languages
- [x] Only device detect fully supported languages
- [x] Render less supported languages differently
- [x] Alert when trying to switch to a less supported language
- [x] Add Spanish as a potential language

| Language switcher | Confirm dialogue
|-|-|
| image | image |


Lisa29k (approved):
Could not have written any better code myself. You got my full support on this one.

PR #962, opened 2022-11-10
Update outro portal with mirror intro portal fallback

Nipher:
Issue: [Notion Task](https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#9d6e226db30c4d899844348a7864867b) ⛑️ 

### TODO

- [x] Check if the video loop sound is reversed and if it sounds good 😬  (SOUNDS SUPER COOL ACTUALLY HAHA tested on Accepting thoughts and feelings since it has video audio for the transition)

### Description of the Change

When outro portal's content isn't defined, the app will take the intro portal and play it backwards. 
As if all ends mirrors their beginnings, ouffffff 🧙 

#### Screenshots

https://user-images.githubusercontent.com/7523828/201306858-c1f9d415-3de0-4907-9ce2-2c948ebdea18.mov

https://user-images.githubusercontent.com/7523828/201307464-44b3414f-614c-4624-88a7-31337a01ea4d.mov

### DIFF
--- a/client/src/routes/Session/OutroPortal.tsx
+++ b/client/src/routes/Session/OutroPortal.tsx
@@ -19,6 +19,7 @@ import {useIsFocused} from '@react-navigation/native';
 import Button from '../../common/components/Buttons/Button';
 import {useTranslation} from 'react-i18next';
 import Gutters from '../../common/components/Gutters/Gutters';
+import AudioFader from './components/AudioFader/AudioFader';
 
 const VideoStyled = styled(VideoBase)({
   ...StyleSheet.absoluteFillObject,
@@ -46,6 +47,7 @@ const TopBar = styled(Gutters)({
 
 const OutroPortal: React.FC = () => {
   const finalVidRef = useRef<Video>(null);
+  const [videoLoaded, setVideoLoaded] = useState(false);
   const exercise = useSessionExercise();
   const {leaveSession} = useLeaveSession();
   const [readyToLeave, setReadyToLeave] = useState(false);
@@ -81,6 +83,7 @@ const OutroPortal: React.FC = () => {
 
   const onEndVideoLoad = () => {
     finalVidRef.current?.seek(0);
+    setVideoLoaded(true);
   };
 
   const onLoopVideoEnd = () => {
@@ -98,6 +101,15 @@ const OutroPortal: React.FC = () => {
         introPortal?.videoLoop?.source &&
         introPortal?.videoEnd?.source && (
           <>
+            {isFocused && introPortal?.videoLoop?.audio && (
+              <AudioFader
+                source={introPortal?.videoLoop.audio}
+                repeat
+                paused={!videoLoaded}
+                volume={readyToLeave ? 1 : 0}
+                duration={readyToLeave ? 20000 : 5000}
+              />
+            )}
             <VideoStyled
               ref={finalVidRef}
               onLoad={onEndVideoLoad}
--- a/client/src/routes/Session/OutroPortal.tsx
+++ b/client/src/routes/Session/OutroPortal.tsx
@@ -20,6 +20,7 @@ import Button from '../../common/components/Buttons/Button';
 import {useTranslation} from 'react-i18next';
 import Gutters from '../../common/components/Gutters/Gutters';
 import AudioFader from './components/AudioFader/AudioFader';
+import Sentry from '../../lib/sentry';
 
 const VideoStyled = styled(VideoBase)({
   ...StyleSheet.absoluteFillObject,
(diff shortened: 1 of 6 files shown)


client/src/routes/Session/OutroPortal.tsx
```diff
@@ -14,43 +14,121 @@ import VideoBase from './components/VideoBase/VideoBase';
...
+import {useIsFocused} from '@react-navigation/native';
+import Button from '../../common/components/Buttons/Button';
+import {useTranslation} from 'react-i18next';
 
 const VideoStyled = styled(VideoBase)({
   ...StyleSheet.absoluteFillObject,
   flex: 1,
 });
 
+const reverseVideo = (url: string) => {
```
Nipher, line 27: need to do some research to see how robust this would be, but pretty neat that cloudnary has `e_reverse` option 🤓

client/src/routes/Session/OutroPortal.tsx
```diff
@@ -14,43 +14,146 @@ import VideoBase from './components/VideoBase/VideoBase';
...
+              <AudioFader
+                source={introPortal?.videoLoop.audio}
+                repeat
+                paused={!videoLoaded}
+                volume={readyToLeave ? 1 : 0}
+                duration={readyToLeave ? 20000 : 5000}
+              />
+            )}
+            <VideoStyled
+              ref={finalVidRef}
```
Nipher, line 114: I'll rename this ref, doesn't make sense to call it `finalVid` in this context

client/src/routes/Session/hooks/useLeaveSession.ts
```diff
@@ -7,21 +7,24 @@ import {NativeStackNavigationProp} from '@react-navigation/native-stack';
...
   const {leaveMeeting} = useContext(DailyContext);
   const {navigate} = useNavigation<ScreenNavigationProps>();
+  const {fetchSessions} = useSessions();
 
   const resetSession = useSessionState(state => state.reset);
 
   const leaveSession = useCallback(async () => {
     await leaveMeeting();
     resetSession();
+    fetchSessions();
```
Nipher, line 25: @gewfy what do you think? I added this cause I noticed that the session that I was in still remained on the sessions list, and I guess it doesn't hurt to refresh it

client/src/routes/Session/OutroPortal.tsx
```diff
@@ -14,43 +14,146 @@ import VideoBase from './components/VideoBase/VideoBase';
...
+    finalVidRef.current?.seek(0);
+    setVideoLoaded(true);
+  };
+
+  const onLoopVideoEnd = () => {
+    ReactNativeHapticFeedback.trigger('impactHeavy');
+    setReadyToLeave(true);
+  };
+
+  console.log(readyToLeave);
```
Nipher, line 94: leftover

client/src/routes/Session/OutroPortal.tsx
Nipher: example:
```
cloudinary.video("ski_jump", {controls:true, transformation: [
    {overlay: "video:ski_jump", flags: "splice", effect: "reverse"},
    {overlay: "video:ski_jump", flags: "splice", effect: "accelerate:-50"},
    {effect: "brightness:10", radius: "max"}
    ]})
```

client/src/routes/Session/hooks/useLeaveSession.ts
gewfy: Agree! Makes sense :)

client/src/routes/Session/OutroPortal.tsx
gewfy: It's something itchy with relying on the field being a cloudinary url when it basically could be anything. That's why I'm twisting and turning about this. But maybe we should go with just a regex to begin with.

Nipher: hmm, I see 🤔  

Are we planning or have indications that we're going to use something different than a cloudnary url on those fields at some point?

I could add a cloudinary check on the regexp as well if we go with that, and I right now if the regexp doesn't match it'll just return the unadulterated url... 

Could also add a hint on the CMS field asking that a cloudnary video asset is used 🤔

Nipher: Might be a good idea to add a sentry log when the regexp "fails" to find a match.

Nipher: [1ca8e87](https://github.com/29ki/29k/pull/962/commits/1ca8e87b5da3268dcf7dd8b0f7e90a6802104e2d)

gewfy: Nice I think it's good for now :)

client/src/routes/Session/OutroPortal.tsx
```diff
@@ -14,43 +14,148 @@ import VideoBase from './components/VideoBase/VideoBase';
...
+
+  if (transformFlags === undefined) {
+    Sentry.captureException(
+      new Error('Could not reverse the video - Invalid url'),
+    );
+    return url;
+  }
+
+  if (transformFlags === '') {
+    return url.replace('/upload/v', '/upload/q_auto,t_global,e_reverse/v');
```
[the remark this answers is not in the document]
Nipher: hmm, strange, if you look into content it seem like a bunch of videos are using that
should we remove it from them as well?

Screenshot 2022-11-15 at 10 19 57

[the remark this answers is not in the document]
Nipher: [e5f4368](https://github.com/29ki/29k/pull/962/commits/e5f4368ff38cbe528cf8c723c8bf1759d687df90)

PR #990, opened 2022-11-14
Editable CMS defaults

gewfy:
Issue: https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#21465b6cfe2240b2b9975c01793c915c

### Description of the Change

This makes it possible to set defaults when creating a new exercise from the CMS. Also adds jest unit testing to the CMS.

#### Screenshots

| New settings collection | Exercise defaults settings |
|-|-|
| image | image |


gewfy:
Maybe useful for you too @Greenheart ? 🤗

Greenheart:
Nice! Thanks for the mention - I'll keep this in mind if we need it in the future! :)

cms/src/lib/fields.ts
```diff
@@ -0,0 +1,37 @@
+import {CmsField} from 'netlify-cms-core';
+import {JSONObject} from '../../../shared/src/types/JSON';
+
+// This applies field defaults to nested CMSField structures as you have to specify them individually per field
+export const applyDefaults: (
```
swemail, line 5: Recursion inside a map to it's outer function, love it!

swemail (approved):

PR #1004, opened 2022-11-15
Add modelUtils to shared and use it on snapshot subscriptions

Nipher:
Issue: [Task in Notion ⛑️ ](https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#adfc7bcb6fd0440dad0b932a0da9b0d6)

### Description of the Change

We transform firestore js documents to an object that matches the `Session` type and we were currently only doing that on the server for API responses, since the server is the main layer that interacts with firestore.

But! we also use firestore `onSnapshot` api to subscribe to live database data, therefore it would be nice to use that code for both use-cases.

This moves that code to `shared` as `modelUtils`

### DIFF
--- a/shared/src/types/Session.ts
+++ b/shared/src/types/Session.ts
@@ -7,22 +7,17 @@ export enum SessionType {
   private = 'private',
 }
 
-// Input to DB
-export type ExerciseStateInput = Omit<ExerciseState, 'timestamp'> & {
+// Data stored in DB
+export type ExerciseStateData = Omit<ExerciseState, 'timestamp'> & {
   timestamp: Timestamp;
 };
 
-export type SessionInput = Omit<Session, 'exerciseState' | 'startTime'> & {
-  exerciseState: ExerciseStateInput;
+export type SessionData = Omit<Session, 'exerciseState' | 'startTime'> & {
+  exerciseState: ExerciseStateData;
   startTime: Timestamp;
   createdAt: Timestamp;
   updatedAt: Timestamp;
 };
-
-// Data stored in DB
-export type ExerciseStateData = ExerciseStateInput;
-
-export type SessionData = SessionInput;
 
 // Applicaton schema
 export type ExerciseState = {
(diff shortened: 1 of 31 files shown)


Nipher:
just realized we also have https://github.com/29ki/29k/blob/a5cca8b08c1efea5a9071f1a31ea8305a973af06/functions/src/controllers/sessions.ts#L12-L14

should that also be run on document snapshot subscriptions?

gewfy (approved):
Greaaat work on this!! 👌 🙌

shared/src/types/Session.ts
```diff
@@ -7,23 +7,18 @@ export enum SessionType {
...
-// Input to DB
-export type ExerciseStateInput = Omit<ExerciseState, 'timestamp'> & {
+// Data stored in DB
+export type ExerciseStateData = Omit<ExerciseState, 'timestamp'> & {
   timestamp: Timestamp;
 };
 
-export type SessionInput = Omit<Session, 'exerciseState' | 'startTime'> & {
-  exerciseState: ExerciseStateInput;
+export type SessionData = Omit<Session, 'exerciseState' | 'startTime'> & {
```
Nipher, line 15: we should probably omit the `hostProfile` field here as well 🤔

Nipher: Ah ok, probably only possible to get the host profile through firebase-admin auth and not the client's...
hm... cause without omitting this prop here we would get inaccurate `Session` types on the client subscriptions (and probably other places) since it actually does not contain `hostProfile`.

swemail: Bummer, don't think we can get it from auth on the client, and that is a good thing, but it messes things up here. Maybe we should actually store the hostProfile on the session document then?

swemail: But if we store it it can get out of sync. Such a pity to have to make a request for that

Nipher: Updated the types and created a separate type ”with host profile" [8426d5c](https://github.com/29ki/29k/pull/1004/commits/8426d5c7e5d3eb6cb69ef92bd135086371bcb232)

PR #1011, opened 2022-11-16
PostHog Metrics lib

gewfy:
- [x] `metrics.setConsent`
- [x] `metrics.logEvent`
- [x] `metrics.setUserProperties`
- [x] `metrics.setCoreProperties`
- [x] Enforce event and user properties types
- [x] Implement a couple of events and user properties
- [x] Updated `INSTALALTION.md`
- [x] Fix tests / mocks
- [x] Setup `production` and `staging` environments in PostHog
- [x] Repository secrets
- [ ] Invite everyone to PostHog


client/src/lib/metrics/types.ts
```diff
@@ -0,0 +1,36 @@
+import {Session} from '../../../../shared/src/types/Session';
+
+export type Events = {
```
Nipher, line 3: types!!! Nice!

shared/src/types/UtilityTypes.ts
```diff
@@ -0,0 +1,9 @@
+// credits goes to https://stackoverflow.com/a/50375286
+// function intersection producec - functin overloads
+export type UnionToIntersection<U> = (
```
Nipher, line 3: 🤯

client/src/lib/metrics/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+// Posthog event user properties https://posthog.com/docs/integrate/user-properties
+type EventUserProperties = {
+  $set?: AnyUserProperty;
+  $set_once?: AnyUserProperty;
+  $unset?: [keyof UserProperties];
+};
+
+// Generate all possible allowed event names and properties as logEvent overloads
+type LogEventOverloads = UnionToIntersection<
```
Nipher, line 49: black magic right there hahaha
nice to get typed stuff on metrics

Nipher (approved):
Dope 🔥

shared/src/types/UtilityTypes.ts
gewfy: Hahaha don't get me started on this 😂 But basically it was a deep ended hole getting a type object structure:
```javascript
 { eventName: { eventProp: type } }
```
Into be type checked in a function where they are separate arguments i.e. `logEvent(eventName, { eventProp: type });`.
What it does in the end is to create multiple function "overloads" of every possible combination:
```typescript
('some event name', { 'some event prop': string}) => void &
('some second event name', { 'some other event prop': boolean }) => void &
('some third event name', { 'some other event prop': boolean }) => void &
```
... and so forth...

Does it make sense? :D

client/src/lib/metrics/index.tsx
gewfy: Thanks! 🤗

client/src/lib/metrics/types.ts
gewfy: Thanks! 🙏

PR #1013, opened 2022-11-17
Show slide content also for hosts

kajsaunge:
Issue: [Host notes is hiding slide content](https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#8a97b82bd9de431988beca230de36b89)

### Description of the Change
- Expand host notes in intro portal
- Push down content for hosts so that text is never hidden

#### Screenshots

| Before | After |
|-|-|
| image | image |

| iOS | Android |
|-|-|
| image | image |

### DIFF
--- a/client/src/routes/Session/Session.tsx
+++ b/client/src/routes/Session/Session.tsx
@@ -7,6 +7,7 @@ import {
   BottomSafeArea,
   Spacer12,
   Spacer16,
+  Spacer32,
   TopSafeArea,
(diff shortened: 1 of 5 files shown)


Nipher:
Would it be possible to use the additional spacer in the `Session` instead of passing it down to the `Content` component? 🤔 💭

Nipher (approved):

PR #1017, opened 2022-11-18
Session mute notification

kajsaunge:
Issue: [Visual ping for hosts when auto playing content](https://www.notion.so/29k/Closed-BETA-0055aabd797b40c7b390b7cea4eba658#b0df34f2c29948a58fdd02ead0023a1b)

### Description of the Change

- Use session notification to show host mute 
- Create session notification state

#### Screenshots

| iOS | Android |
|-|-|
| image | image |


gewfy (approved):
Nice!! :)

client/src/routes/Session/state/sessionNotificationsState.ts
```diff
@@ -0,0 +1,24 @@
+import create from 'zustand';
+import {NotificationProps} from '../components/Notifications/Notification';
+
+type State = {
+  notifications: NotificationProps[];
```
gewfy, line 5: Nit pick: Like this reused type, but would maybe do it in reverse. Basically define what a notification is in this file, and then use that definition for the props in the component :)

kajsaunge: 100%

PR #1022, opened 2022-11-21
Move type closer to definition

kajsaunge:
Move type to state to be reused by component


client/src/routes/Session/state/sessionNotificationsState.ts
```diff
@@ -1,5 +1,14 @@
 import create from 'zustand';
-import {NotificationProps} from '../components/Notifications/Notification';
+import {IconType} from '../../../common/components/Icons';
+
+export type NotificationProps = {
```
gewfy, line 4: How about just calling `Notification`? :)

kajsaunge: Fixed 🌭

gewfy (approved):

PR #1055, opened 2022-11-22
Verification code improvements

swemail:
Verification code improvements according to design

drawing
drawing
drawing


gewfy (approved):
Nice LGTM!

PR #1096, opened 2022-11-25
Prevent state race condition on host notes navigation

Nipher:
Issue: #1095 

### Description of the Change

~Adds a `return` statement to prevent the `onPress` execution.~

Re-actively navigates to slides based on scroll state.

### Verification Process

If you try to manually press the host note navigation buttons faster than the `activeIndex` state  updates it would cause the error, now, with it being reactive to the state, the issue should not happen anymore.


swemail (approved):
Nice!

PR #1133, opened 2022-11-30
Add q_auto:eco to video cloudnary urls

Nipher:
Issue: https://29k.slack.com/archives/C7DTMLM6Y/p1669768023205229

Below you can see the `q_auto:eco` being `8.1mb`, the normal version as `15.4mb` and the `q_auto` version with `10.6mb`.

Screenshot 2022-11-30 at 07 51 57


gewfy:
Don't forget to change the cms presets https://github.com/29ki/29k/blob/main/cms/src/fields/constants.ts

kajsaunge (approved):

PR #1144, opened 2022-12-01
Avoid accidental text selection

kajsaunge:
Set selectable false on buttonText

### DIFF
--- a/client/src/common/components/Buttons/Button.tsx
+++ b/client/src/common/components/Buttons/Button.tsx
@@ -37,7 +37,7 @@ type ButtonTextProps = {
   children: React.ReactNode;
 };
 
-const ButtonText = styled(Body16)<ButtonTextProps>(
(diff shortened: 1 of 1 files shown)


kajsaunge:
Adding an attribute results in a type complaint, can you help me understand @swemail ?

swemail:
> Adding an attribute results in a type complaint, can you help me understand @swemail ?

It's because the `Body16` has the `selecteble: true` attribute and that is not possible to override it seems. Don't know the best solution here

swemail:
> Adding an attribute results in a type complaint, can you help me understand @swemail ?

Maybe just create a separate `Body16` component just for this one?

swemail (approved):

kajsaunge:
> It's because the `Body16` has the `selecteble: true` attribute and that is not possible to override it seems. Don't know the best solution here

Thanks for explaining this Mr Type 🙏

PR #1145, opened 2022-12-01
Throw errors when running functions locally

gewfy:
Had silent auth issues (again) when switching between local and remote firebase environments. Can't believe we didn't add this earlier :P 

image


swemail (approved):
🥳

PR #1148, opened 2022-12-01
Makes it possible to hide published content and toggle to show in staging and dev

swemail:
This is so that publicHosts can switch to see all content in staging and dev

<image src="https://user-images.githubusercontent.com/1139207/205051212-ddde942e-2a29-41f3-a119-6743644646da.png" width="200" /> <image src="https://user-images.githubusercontent.com/1139207/205051357-fa32993b-4c20-4211-92b3-eaf7bd945aad.png" width="200" />


client/src/lib/content/hooks/useExerciseIds.ts
```diff
@@ -1,6 +1,30 @@
...
+  const ids = exerciseIds as Array<string>;
+
+  const onlyPublished = useMemo(() => {
+    const pred = (id: string) => {
+      // @ts-expect-error variable/string litteral as key is not yet supported https://www.i18next.com/overview/typescript#type-error-template-literal
+      const exercise = t(id, {
+        returnObjects: true,
+      }) as Exercise;
+
+      return exercise?.published === true;
```
gewfy, line 21: I vote for us actually having published and "hidden" or "wip" as two different things. The reason is that we wan't languages to be able to fallback to english if it's not available yet, but maybe under translation process.

gewfy (commented):
My 🧠 is a bit too mushy, so this review might be off. But I would like for us to fully understand the editors need (e.g. Jenny) in combination with how we deal with language fallback before.

gewfy (changes requested):
Blocking this for now :)

cms/src/fields/common.ts
```diff
@@ -22,6 +22,17 @@ export const PUBLISHED_FIELD: CmsField = {
...
+};
+
+export const HIDDEN_FIELD: CmsField = {
+  label: '🙈 Hidden',
+  name: 'hidden',
+  widget: 'boolean',
+  required: false,
+  default: false,
+  i18n: true,
+  hint: "This will make it hidden by default. For work in progress, use in combination with published and the 'Show work in progress' switch in the app to access it only in staging.",
```
gewfy, line 35: 👌 👌 👌 👌

client/src/lib/appState/hooks/useToggleHiddenContent.ts
```diff
@@ -0,0 +1,24 @@
...
+      LANGUAGE_TAGS.forEach(lng => i18n.removeResourceBundle(lng, 'exercises'));
+      i18n.reloadResources(undefined, 'exercises');
+    },
+    [i18n, setShowHiddenContent],
+  );
+
+  return toggleHiddenContent;
+};
+
+export default useToggleHiddenContent;
```
gewfy, line 24: Could we move this to `i18n` lib instead?

client/src/lib/i18n/index.ts
```diff
@@ -18,20 +18,39 @@ import {
...
+const omitExercises = (resources: typeof content.i18n) => {
+  const allResources = clone(resources) as unknown as Record<
+    string,
+    Record<string, string>
+  >; // Can't delete exercises in type content.i18n since it's not opttional
+  for (const ln of LANGUAGE_TAGS) {
+    delete allResources[ln].exercises;
+  }
+  return allResources;
+};
```
gewfy, line 41: Nit pick, how about moving this function and `useAppState.getState().showHiddenContent` to a utils file?

gewfy (approved):
Really great! So much simpler imo 🙌  None of my comments are hard stops. So merge away if you don't agree 😄

client/src/lib/appState/hooks/useToggleHiddenContent.ts
swemail: Right, forgot in my frustration, will do

client/src/lib/i18n/index.ts
swemail: Will do

PR #1163, opened 2022-12-02
Refactor useSessionExercise to more specific useSessionSlideState

Nipher:
### Description of the Change

The `useSessionExercise` hook did more than just providing the session exercise, it calculated which slide is the `previous`, `current` and `next`, which results in a new object being created for every time the session exercise state index changes.

This change removes the rest of the exercise information from the value it returns, so we don't use this hook for those, that data is already available and is independent from the current session state index, meaning it won't change.

So this change removes the potential of using a hook to access data in a way that may  trigger more re-renders than necessary.

And with that I also decided to name it `useSessionSlideState`, let me know if you have better names in mind for it.


gewfy (approved):
Nice! Wouldn't it be nice to have a `useSessionExercise`too? That simplifies this...
```
const exercise = useExerciseById(session?.contentId);
const session = useSessionState(state => state.session);
```

PR #1169, opened 2022-12-06
Add re-edit session type

Nipher:
Issue: [Notion Task ⛑️ ](https://www.notion.so/29k/Early-Access-2794500652b34c64b0aff0dbbc53e0ab#52a1b76699e546d693178736b84b614b)

- [x] Add it to create session flow
- [x] Edit existing session flow

### Description of the Change

Add session type copy and re-edit functionality.

#### Creating a session

Session type icon on the video below is not updating according to the type, [e57f3c5](https://github.com/29ki/29k/pull/1169/commits/e57f3c51b634cf1113aa0c281cefdab6dfafe7e9) fixes that 😬 

https://user-images.githubusercontent.com/7523828/205902112-80a71002-4d98-4fac-bfce-34dbdfafcbfd.mp4

#### Editing an existing session

https://user-images.githubusercontent.com/7523828/205930791-0455d921-2c1f-4412-9d6a-f3903e0d5384.mp4

### DIFF
--- a/client/src/routes/CreateSessionModal/components/steps/SelectTypeStep.tsx
+++ b/client/src/routes/CreateSessionModal/components/steps/SelectTypeStep.tsx
@@ -1,0 +1,131 @@
+import React, {useMemo} from 'react';
+import {useTranslation} from 'react-i18next';
+import styled from 'styled-components/native';
+import {COLORS} from '../../../../../../shared/src/constants/colors';
+import {SessionType} from '../../../../../../shared/src/types/Session';
+import Byline from '../../../../common/components/Bylines/Byline';
+import Gutters from '../../../../common/components/Gutters/Gutters';
+import {PrivateIcon, PublicIcon} from '../../../../common/components/Icons';
+import Image from '../../../../common/components/Image/Image';
+import {
+  Spacer16,
+  Spacer28,
+  Spacer8,
+} from '../../../../common/components/Spacers/Spacer';
+import TouchableOpacity from '../../../../common/components/TouchableOpacity/TouchableOpacity';
+import {Body16} from '../../../../common/components/Typography/Body/Body';
+import {Display24} from '../../../../common/components/Typography/Display/Display';
+import {ModalHeading} from '../../../../common/components/Typography/Heading/Heading';
+import {SPACINGS} from '../../../../common/constants/spacings';
+import useExerciseById from '../../../../lib/content/hooks/useExerciseById';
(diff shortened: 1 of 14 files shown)


client/src/routes/SessionModal/SessionModal.tsx
```diff
@@ -122,11 +165,18 @@ const SessionModal = () => {
       'Session Type': session.type,
       'Session Start Time': session.startTime,
     });
-  };
+  }, [
+    session.startTime,
```
swemail, line 169: If sessions does not change much using the whole object as dependency is cleaner IMHO

client/src/routes/SessionModal/SessionModal.tsx
```diff
@@ -138,9 +188,18 @@ const SessionModal = () => {
       'Session Type': session.type,
       'Session Start Time': session.startTime,
     });
-  };
+  }, [
+    exercise?.name,
```
swemail, line 192: Same as above, thinks it's ok to depend on `session` and `exercise` here

swemail (approved):

client/src/routes/CreateSessionModal/components/steps/SelectTypeStep.tsx
```diff
@@ -0,0 +1,131 @@
...
+              nextStep();
+            }}
+            label={t(`selectType.${type}.title`)}
+            Icon={type === 'private' ? <PrivateIcon /> : <PublicIcon />}
+          />
+          {i < arr.length - 1 && <Spacer16 />}
+        </TypeItemWrapper>
+      )),
+    [t, nextStep, setSelectedType],
+  );
```
gewfy, line 99: This is interesting! I had the same issue with how to deal with callbacks in maps. Like it :)

gewfy (approved):
Nice :)

client/src/routes/SessionModal/SessionModal.tsx
Nipher: hmmm... not sure, like if we set the session state to a new object but it is still the same session to a large extent (fields we might depend upon didn't change), it will trigger unnecessary code, no?

client/src/routes/SessionModal/SessionModal.tsx
Nipher: Yeah exercise it actually makes no sense to not use the whole thing as a dependency since useExerciseById memoizes the whole content by id already

PR #1237, opened 2022-12-09
Show session duration

swemail:
Show duration on session cards 
<image src="https://user-images.githubusercontent.com/1139207/206686417-2a962ffb-7a12-4c51-acdb-c89f136b2fae.png" width="200"> <image src="https://user-images.githubusercontent.com/1139207/206687498-951aca75-f0a2-40b2-bae2-7e21cd561ee8.png" width="200"> <image src="https://user-images.githubusercontent.com/1139207/206687050-73cd8249-bd40-4af6-b5d7-562a592e0eb7.png" width="200">


Nipher (approved):
Nice! I guess we wouldn't want to use `numberOfLines={1}`, right? 
since some names could be long enough to have two lines....

swemail:
> Nice! I guess we wouldn't want to use `numberOfLines={1}`, right? since some names could be long enough to have two lines....

Yea, exactly, we wont it to wrap. Preferable the duration part would never wrap but couldn't figure that out so I left it like this for now

PR #1291, opened 2022-12-14
Only recreate user when getIdToken fails with revoked or expired

swemail:
Flip recreation of user on expected errors revoked or expired when fetching id token

@gewfy we swallow all promise rejections so the user has no way of knowing there was an error, should we fix that now?

### DIFF
--- a/client/src/lib/apiClient/apiClient.spec.ts
+++ b/client/src/lib/apiClient/apiClient.spec.ts
@@ -66,10 +66,10 @@ describe('apiClient', () => {
     });
   });
 
-  it('recreates user if authorization header from server fails', async () => {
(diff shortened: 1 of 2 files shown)


gewfy (approved):

swemail:
> Do yo
> 
> > Flip recreation of user on expected errors revoked or expired when fetching id token
> > @gewfy we swallow all promise rejections so the user has no way of knowing there was an error, should we fix that now?
> 
> Do you have an idea on how to prompt the user about it? I'm clueless 🤔

Not really while apiClient is outside React, now for each place we use it we need handle it so that we can set some state to show a retry or something

PR #1302, opened 2022-12-15
Move session "live" data into subcollection

transient-jonas:
Issue: [Notion Task 🪖 ](https://www.notion.so/29k/Early-Access-2794500652b34c64b0aff0dbbc53e0ab#a72ea84d51ac470ba31dee365b5f710a)

### Description of the Change

- [x] Do a POC with the concept/structure locally and figure out if it works
  - [x] Renaming
  - [x] Specs
  - [x] Set current session on zustand together with sessionState
  - [x] Manual test
- [x] Make it client backwards compatible ? (or probably just killswitch)

#### Migrate existing data to new format 

- [ ] Go to  #1309
  - [x] Use script against staging or testing to verify
  - [ ] Use script against production
  - [ ] Deploy #1302 (this PR)
  - [ ] Run script again in case some update happened during deploy? (probably not since updates only happen while sessions are in progress)

- [x] Discuss sessions query filters (current solution: `ended: true` is set on session as well [ff62e38](https://github.com/29ki/29k/pull/1302/commits/ff62e38854eb054c4d6b571219467d67577494d2)) Good enough solution IMO


client/src/routes/Session/hooks/useSubscribeToSession.ts
```diff
@@ -1,22 +1,43 @@
...
+      onSnapshot: (
+        sessionState: SessionState | undefined,
+        session: Session | undefined,
+      ) => any,
+    ) => {
+      const sessionDoc = firestore().collection('sessions').doc(sessionId);
+      const stateDoc = sessionDoc.collection('state').doc(sessionId);
 
-      const unsubscribe = doc.onSnapshot(
+      const sessionPromise = sessionDoc.get();
```
transient-jonas, line 26: We can replace this by an API call too if we want.

gewfy:
@transient-jonas FYI we decided to wait with this PR til we're back from holidays. Less risk of messing up while we're away :)

client/src/lib/uiLib/components/DrawerMenu.tsx
```diff
@@ -58,6 +58,7 @@ const DrawerContent: (
     const navState = navigation.getState();
     const activeRouteName = navState.routes[navState.index].name;
 
+    /* eslint-disable react/no-unstable-nested-components */
```
swemail, line 61: 👌

client/src/routes/Session/Session.tsx
```diff
@@ -128,23 +127,23 @@ const Session = () => {
...
+    if (sessionState?.completed) {
       logSessionMetricEvent('Complete Sharing Session');
     }
-  }, [session?.exerciseState?.completed, logSessionMetricEvent]);
+  }, [sessionState?.completed, logSessionMetricEvent]);
 
   useEffect(() => {
-    if (session?.ended) {
+    if (sessionState?.ended) {
       leaveMeeting();
```
swemail, line 143: this is async so maybe await it and then navigate. Could return to something unmounted otherwise or perhaps keep it in memory

client/src/routes/Session/hooks/useSubscribeToSession.ts
swemail: Guess not, we a complete firestore doc to be able to subscribe right? What comes from an api call is just data

client/src/routes/Session/hooks/useUpdateSessionState.ts
```diff
@@ -0,0 +1,83 @@
...
+  const setPlaying = useCallback(
+    async (playing: SessionState['playing']) => {
+      if (sessionId) {
+        return sessionApi.updateSessionState(sessionId, {playing});
+      }
+    },
+    [sessionId],
+  );
+
+  return useMemo(
```
swemail, line 66: Don't think we need memo here, if we always use it like `const {setPlaying} = useUpdateSessionState()` but maybe we don't always do? If we somewhere use it like `const updateSessionState = useUpdateSessionState()` memo is needed

client/src/routes/Session/hooks/useSubscribeToSession.ts
gewfy: But this is the session part that we don't subscribe to. This should def. be fetched with rest imo.

gewfy: Or fetch the session separate from this hook, since it's actually not a subscription

transient-jonas: I'll try the navigation param approach 👍

PR #1309, opened 2022-12-16
Add migrate session state script

transient-jonas:
ref #1302


transient-jonas:
The script works, just had to do a minor change so we don't migrate unnecessary data 5adbdc0

gewfy (approved):

PR #1362, opened 2022-12-22
Log setting interested on session

swemail:
Refactor to using hook for logging events around but not inside sessions.

@gewfy should we use the same pattern for logging inside sessions, with the hook accepting session as argument instead of using it from state?


gewfy (approved):
I guess this one is ready to be released? :)

PR #1365, opened 2022-12-22
Persist show hidden content

swemail:


client/src/lib/appState/state/state.ts
```diff
@@ -44,7 +44,10 @@ const useAppState = create<State & Actions>()(
     {
       name: 'appState',
       getStorage: () => AsyncStorage,
-      partialize: ({settings}) => ({settings}),
+      partialize: ({settings, showHiddenContent}) => ({
+        settings,
+        showHiddenContent,
```
gewfy, line 49: How about moving this into `settings` instead? :)

swemail: 👍

gewfy (approved):

PR #1389, opened 2022-12-27
Fix issue with modals being clipped on input focus

gewfy:
Fixes issues where sheet modals gets clipped when focusing/bluring inputs on android. Also see https://github.com/gorhom/react-native-bottom-sheet/issues/618

# Before
https://user-images.githubusercontent.com/474066/209643082-1358a9c7-0afc-41a2-b33d-4b45ce565eac.mov

# After
https://user-images.githubusercontent.com/474066/209642847-77ac1826-df02-410e-8b92-6a7718263eb6.mov


Greenheart:
Awesome! Thank you 🥳🙌

PR #1402, opened 2022-12-28
Better abstraction for checking and alerting about app permissions

gewfy:
- [x] Re-use hook in `ChangingRoom` and `Session`
- [x] Restart JS-bundle after opening settings to re-evaluate permissions - haven't found a better solution on Android, iOS does this by default
- [x] Tests


swemail (approved):
Splendid!

PR #1487, opened 2023-01-11
Validate user for private sessions

swemail:
Validate that a user is part of a private session when requesting a token. Handle errors in the client for when user is not part or session is not found. None of witch should happen from regular usage.

This should be safe to merge into main and not need to wait for a release, will only affect clients that requests tokens

### DIFF
--- a/client/src/routes/Session/ChangingRoom.tsx
+++ b/client/src/routes/Session/ChangingRoom.tsx
@@ -6,7 +6,7 @@ import {
   useRoute,
 } from '@react-navigation/native';
 import {NativeStackNavigationProp} from '@react-navigation/native-stack';
-import {ActivityIndicator, Platform} from 'react-native';
+import {ActivityIndicator, Alert, Platform} from 'react-native';
 import styled from 'styled-components/native';
 import {useTranslation} from 'react-i18next';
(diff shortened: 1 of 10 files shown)


client/src/routes/Session/ChangingRoom.tsx
```diff
@@ -138,20 +138,44 @@ const ChangingRoom = () => {
   const hasVideo = Boolean(me?.videoTrack);
 
   useEffect(() => {
-    if (session?.id) {
+    // If switching between sessions, the session will first be the old one
```
transient-jonas, line 141: Does switching between sessions means leaving a session and entering another one through the UI?
I thought we cleared the session state rather than keeping the old session on 🤯

transient-jonas (approved):

client/src/routes/Session/ChangingRoom.tsx
swemail: Found this when going back from the `ChangingRoom` for different sessions, so it has probably not been happening too much but still, better to be a 100%

swemail: So not when leaving the regular way

PR #1490, opened 2023-01-11
Use owner instead of dailySpotlightId

swemail:
Use owner from daily to determine if the user is host, this will require all users to be using tokens


gewfy (approved):
its-amazing-impressed

PR #1495, opened 2023-01-12
Host mutes all when playing non sharing

swemail:
Since tokens make it possible for the host to mute all the host will send mute to all on playing non sharing exercises instead of each client muting itself on state changes


client/src/routes/Session/hooks/useMuteAudio.ts
```diff
@@ -0,0 +1,32 @@
...
+  );
+
+  const conditionallyMuteParticipants = useCallback(
+    (playing: boolean, currentSlide: ExerciseSlide | undefined) => {
+      if (playing && currentSlide?.type !== 'sharing') {
+        muteAll();
+        addNotification({
+          text: t('notifications.muted'),
+          Icon: MicrophoneOffIcon,
+        });
```
gewfy, line 23: I believe this notification will only be shown for the host that mutes. Maybe we need to move this logic into `SessionNotifications` instead?

gewfy (approved):
Nice! Minor notification thing

client/src/routes/Session/hooks/useMuteAudio.ts
swemail: In the old `useMuteAudioListener` it only adds notification if you're the session host, so I guess that never was shown to the other participants

gewfy: Oh then it was a bug. We can fix that separately. Can you write a todo?

gewfy: Or actually it wasn’t a bug since it was the clients themselves that muted because of the state change. So it would turn up on all clients.

swemail: Tested it now without my change, only the host got the notification

swemail: Even if the clients muted themselves only the condition for the notification was if you where the session host

swemail:
@gewfy moved notifications to a where it fits better, needed to solve the syncing of updates in a suboptimal way but couldn't figure out anything better. Now everyone gets notifications about being muted

gewfy (approved):

PR #1496, opened 2023-01-12
Dependabot/npm and yarn/client/zustand 4.2.0

swemail:
After some back and forth with the maintainers of zustand this is now ready to be upgraded


gewfy (approved):
Interesting what was the original issue? Is there a GitHub issue in their repo?

swemail:
> Interesting what was the original issue? Is there a GitHub issue in their repo?

https://github.com/pmndrs/zustand/issues/1505, seems picky about how to mock async storage. But I actually think we have had warnings coming from that before.

PR #1501, opened 2023-01-13
Adds tags to cms and card

swemail:
Adds possibility to add tags in the cms and set them on exercises
<image src="https://user-images.githubusercontent.com/1139207/212350059-074598a0-d51e-436c-bd71-7101405b5832.png" width="200" />


client/src/lib/components/Bylines/Byline.tsx
```diff
@@ -30,23 +30,25 @@ type BylineProps = {
...
-  );
-};
+    return (
+      <Container>
+        <ImageContainer>
+          {pictureURL && <Image source={{uri: pictureURL}} />}
+        </ImageContainer>
+        <Spacer4 />
+        <WrapText numberOfLines={2}>
+          {t('with')} {name} {duration ? `· ${duration} min` : ''}
```
gewfy, line 47: Not your change. But `min` needs to be translated

client/src/lib/content/hooks/useTagsById.ts
```diff
@@ -0,0 +1,17 @@
...
+  return useCallback(
+    (id: any) =>
+      t(id, {
+        returnObjects: true,
+      }) as Tag,
+    [t],
+  );
+};
+
+export default useGetTagById;
```
gewfy, line 17: Nit pick, rename file to `useGetTagById` (singular)

client/src/lib/components/Tags/Tags.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <Wrapper key={tag.id}>
+          <Tag key={tag.id} value={tag.tag} />
+          <Spacer4 />
+        </Wrapper>
+      ))}
+    </Container>
+  );
+});
+
+export default Tags;
```
gewfy, line 53: This component is very coupled to the `SessionCard` imo. So I would move it there.

client/src/lib/components/Cards/Card.tsx
```diff
@@ -97,51 +105,51 @@ export const Card: React.FC<CardProps> = ({
...
-    </Wrapper>
-  </>
-);
+}) => {
+  return (
+    <>
+      <Wrapper onPress={onPress}>
+        <ContentWrapper>
+          <LeftCol>
+            <Tags tagIds={tagIds} duration={duration} />
```
gewfy, line 114: Me and Kajsa have tried to make the `Card` component very "stupid" and agnostic. Basically that `SessionCard` is the container that resolves all the data that `Card` need to render.
Here we resolve the tags in this component. Imo I would resolve the complete translated tags in `SessionCard` and just pass that as an array that you just render with `Tag` here.
The filtering logic of tags in `Tags` could be moved to a hook specific to `SessionCard` or as a more agnostic `useGetTagsByIds` in `content/`.

client/src/lib/components/Tags/Tags.tsx
gewfy: Or refactor it to a hook as mentioned ⬆️

cms/src/collections/collections.ts
```diff
@@ -48,6 +49,23 @@ export const settings: CmsCollection = {
...
+  slug: '{{id}}',
+  summary: '{{fields.tag}}',
+  extension: 'json',
+  create: true,
+  format: 'json',
+  delete: true,
+  publish: true,
+  i18n: true,
+};
+
```
gewfy, line 68: Nice 👌

cms/src/fields/tag.ts
```diff
@@ -0,0 +1,21 @@
...
+  ID_FIELD,
+  {
+    name: 'tag',
+    label: '🏷 Tag',
+    widget: 'string',
+    i18n: true,
+  },
+  {
+    label: '🔢 Featured order',
+    name: 'featuredOrder',
```
gewfy, line 14: Nit pick, can we call it just `order`?

gewfy (changes requested):

client/src/lib/components/Tags/Tags.tsx
swemail: Did both 😄

gewfy (approved):
I would have made the tags even more abstract. Basically a an array of strings passed to `Card` where the duration is resolved and added from `SessionCard`. But I leave that up to you. Might change it my self after merge :)

swemail:
> I would have made the tags even more abstract. Basically a an array of strings passed to `Card` where the duration is resolved and added from `SessionCard`. But I leave that up to you. Might change it my self after merge :)

I'll do now then 😂

PR #1538, opened 2023-01-16
Fix useSessionStartTime

transient-jonas:
Issue: [Notion Task 🧮](https://www.notion.so/29k/Early-Access-2794500652b34c64b0aff0dbbc53e0ab#a867277406884164aeef5ed639d41d67)

### Description of the Change

- 6ba24a1 fixes `After the interval starts it never stops`
- d3a4d08 fixes re-renders on the whole portal, now it is only on the status
- cb1fb74 fixes re-renders on the whole session card, now it is only in its button


gewfy (approved):
👍 👍 👍  Maybe remove support for `buttonText` and `onButtonPress` in `Card` also?

transient-jonas:
> 👍 👍 👍 Maybe remove support for `buttonText` and `onButtonPress` in `Card` also?

YESS true! will dooo 🙌

PR #1540, opened 2023-01-17
Minor Card and SessionCard refactor

gewfy:
@swemail this is what I proposed. Does it make sense? :)


swemail (approved):
Make sense, sometimes better to tell in code right 😄

PR #1541, opened 2023-01-17
Include hashed user id in session tokens

swemail:
WAIT FOR #1535 and bump of killswitch!!!!

Include a hash based in room id and user id (+ DAILY_ID_SALT) in the session tokens. When user id is not included in tokens the `user_id` and `session_id` are the same on a daily participant. When user id is included in the tokens the `user_id` and `session_id` are not the same. `active-speaker-change` sends what it calls `peerId` witch is the same as `session_id` so the participant sort order have to find the user based on the `session_id` and therefor setting the order is moved into `dailyState` where the current state can be accessed. Since `session_id` is ephemeral using that for sort order is not stable.


functions/src/lib/id.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import config from './config';
+
+const {DAILY_ID_SALT} = config;
+
+const uuidv5Bytes = (name: string) => uuidv5(name, DAILY_ID_SALT, []);
+const createId = compose(bs58.encode, uuidv5Bytes);
+
+export const createRctUserId = (roomId: string, userId: string) =>
+  createId(`room:${roomId}:user:${userId}`);
```
gewfy, line 13: Nit pick, but what is `rct`? Wouldn't `daily` make more sense?

gewfy (approved):
Nice nice nice nice!

functions/src/lib/id.ts
swemail: Haha, yes, reused the mighty Örjans naming convention. Make kind of sense, they migrated video platform and maybe we will too 😄

swemail: rct, `real time communication` apparently

PR #1594, opened 2023-01-23
Support for capturing interested count

swemail:
Backend/Client support for capturing interested count


transient-jonas (approved):
niiice 🔥

PR #1609, opened 2023-01-24
Completed sessions

transient-jonas:
Issue: [Notion Task ⛑️](https://www.notion.so/29k/Early-Access-2794500652b34c64b0aff0dbbc53e0ab#7030ff39ac6046ad9aca8ff3d360fe86)

### Description of the Change

We are about to create a new app tab with the user's "Plan", displaying completed sessions and more, so we need to know which sessions the user have participated in to completion.

This change is storing those on async storage, for now it only stores the Session ID, perhaps we should also store a date of when it got completed?

Next steps will be to create that new Tab and display and if necessary fetch more info on those sessions.

#### Video

This video shows the completed session being added on the user's local storage.

https://user-images.githubusercontent.com/7523828/214304262-7c300630-3a4b-4d05-80d7-fcf406fb3c50.mp4

### DIFF
--- a/client/src/lib/user/state/state.ts
+++ b/client/src/lib/user/state/state.ts
@@ -11,6 +11,7 @@ type PinnedSession = {
 
 type CompletedSession = {
   id: string;
+  completedAt: Date;
 };
 
 type UserState = {
--- a/client/src/lib/user/state/state.ts
+++ b/client/src/lib/user/state/state.ts
@@ -9,7 +9,14 @@ type PinnedSession = {
   expires: Date;
 };
 
-type UserState = {pinnedSessions: Array<PinnedSession>};
+type CompletedSession = {
+  id: string;
+};
+
+type UserState = {
+  pinnedSessions: Array<PinnedSession>;
+  completedSessions: Array<CompletedSession>;
+};
 
 type State = {
   user: FirebaseAuthTypes.User | null;
@@ -25,6 +32,7 @@ type Actions = {
     claims: State['claims'];
   }) => void;
   setPinnedSessions: (pinnedSessions: Array<PinnedSession>) => void;
+  setCompletedSessions: (completedSessions: Array<CompletedSession>) => void;
   reset: (isDelete?: boolean) => void;
 };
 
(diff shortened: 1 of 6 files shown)


client/src/lib/user/state/state.ts
```diff
@@ -9,7 +9,14 @@ type PinnedSession = {
   expires: Date;
 };
 
-type UserState = {pinnedSessions: Array<PinnedSession>};
+type CompletedSession = {
+  id: string;
```
transient-jonas, line 13: @gewfy @swemail as a first step I was just trying to store something (id)

Should we also add `completedAt`? any other field you see being important in this case? 🤔

swemail: Don't we need to lookup sessions anyway by id? We would have all the data then anyways

gewfy: `completedAt` is definitely needed to create a timeline :P

swemail: It is but I just thought about needing to look it up anyways, we don't know who the host was an such

swemail (approved):
Nice!!!

client/src/lib/user/state/state.ts
transient-jonas: I think I'll add `completedAt` in case we decide to display/fetch only for e.g "the last 3 or so". 
Then we could order/filter that first and gradually fetch it?

client/src/routes/Session/Session.tsx
```diff
@@ -134,7 +139,10 @@ const Session = () => {
 
   useEffect(() => {
     if (sessionState?.completed) {
-      addCompletedSession({id: sessionState?.id});
+      addCompletedSession({
+        id: sessionState?.id,
+        completedAt: dayjs().toDate(),
```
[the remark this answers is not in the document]
transient-jonas: Done [dfe80bc](https://github.com/29ki/29k/pull/1609/commits/dfe80bcd77128cb41bae82a4fffbfeed348d425d)

client/src/lib/user/state/state.ts
transient-jonas: done [aa29319](https://github.com/29ki/29k/pull/1609/commits/aa293194a5a5500c3f77622f0c2ebcffbacee477)

PR #1624, opened 2023-01-26
Improve portal video transition

gewfy:
This improves the intro and outro portal video transitions by isolating the render updates. Also shared `VideoTransition` component used by both `IntroPortal` and `OutroPortal`.

Idea for next iteration would be to show and hide video components with re-animated.


swemail (approved):
Nice, interesting to see how it performs 😄

PR #1634, opened 2023-01-27
Async routes and state

swemail:
Draft!


gewfy (changes requested):
Great start! So nice to see that we can reuse. I think it would be nice if we split this up into `routes/LiveSession` and `routes/AsyncSession` and then move all shared stuff to a new `lib/session`
* I think it more clearly shows where we draw the line for an abstraction
* Also making the Session folder a bit lighter
* `live` instead of `online` since this is what Lisa is referring too in her concepts.

Only opinions, so please take them as ideas. Except it being called `live` maybe :p

PR #1664, opened 2023-01-30
Use common intro portal

swemail:
Make use of a common `IntroPortalComponent` and resolve Live/Async specific things in container components


transient-jonas (approved):
LGTM! Nice work!

client/src/lib/session/hooks/useLogInSessionMetricEvents.ts
```diff
@@ -15,37 +15,65 @@ type AllowedSharingEvents =
...
+        metrics.logEvent(event, {
+          'Sharing Session ID': asyncSession.id,
+          'Sharing Session Type': asyncSession.type,
+          'Sharing Session Start Time': asyncSession.startTime,
+          'Sharing Session Duration': dayjs().diff(
+            asyncSession.startTime,
+            'seconds',
+          ),
+          'Exercise ID': asyncSession.contentId,
+          Host: false,
```
gewfy, line 61: This is a matter of definition 😄  Also I'm thinking of adding a prop that differentiate these two. Like `Sharing Session Mode`: `live/async`. Nothing you need to change. Just thinking loud here... in text... 🤔

gewfy (approved):
👌

client/src/lib/session/hooks/useLogInSessionMetricEvents.ts
swemail: Thought that we still pass sessionType so from that we can deduce that Host is not important for async sessions. But room for improvement, definitely

PR #1666, opened 2023-01-31
Split session to work for live and async

swemail:


transient-jonas (approved):
Nice work splitting things up!

PR #1670, opened 2023-02-01
Async session

swemail:
Makes session compatible with async


gewfy (approved):
Really coming together... I can almost see it infront of me through all the code!

PR #1687, opened 2023-02-02
Add GET /user/:id endpoint

transient-jonas:
### Description of the Change

I need this endpoint to fetch user data for locally (device) stored completed sessions.
This way we do not store any data owned by the user without them being able to delete it and dis-enable others to access such.

### DIFF
--- a/functions/src/api/user/index.ts
+++ b/functions/src/api/user/index.ts
@@ -1,12 +1,17 @@
 import * as yup from 'yup';
 import validator from 'koa-yup-validator';
+
 import {createApiRouter} from '../../lib/routers';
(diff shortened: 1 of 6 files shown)


functions/src/api/user/index.ts
```diff
@@ -78,4 +83,21 @@ userRouter.put(
   },
 );
 
+userRouter.get('/:id', async ctx => {
+  try {
+    const userProfile = await getProfile(ctx.params.id);
```
gewfy, line 88: And maybe a new `getHost` that checks that the user actually has that claim before returning any data

transient-jonas: That was my initial approach, but then I remembered that there are private sessions 😬

PR #1691, opened 2023-02-02
Merge sessionNotification state into session state

gewfy:
Removes the `sessionNotification` state. Will wait for #1632 to not cause unnecessary merge conflicts


swemail (approved):
Great, and good if we wait until #1632 but now it's approved so you can merge later

PR #1707, opened 2023-02-06
Remove duplicate UUID packages

gewfy:
Removes `react-native-uuid` in favour of `react-native-get-random-values`


transient-jonas:
Weird, where does that [skip](https://github.com/29ki/29k/pull/1707/files#file-client-src-lib-content-hooks-usegetexercisesbytype-spec-ts-L43) comes from?

transient-jonas (approved):

gewfy:
> Weird, where does that [skip](https://github.com/29ki/29k/pull/1707/files#file-client-src-lib-content-hooks-usegetexercisesbytype-spec-ts-L43) comes from?

@transient-jonas Oh good catch! I'll have a look!

PR #1708, opened 2023-02-06
Set default fill on SVG icons

gewfy:
Default fill color for SVGs was [restored to black](https://github.com/software-mansion/react-native-svg/pull/1947) in [`react-native-svg@13.7.0`](https://github.com/software-mansion/react-native-svg/releases/tag/v13.7.0)

Mostly noticeable on checkmark, arrow left and logotype that has non-closed paths:
| Before | After |
|-|-|
| image | image |

### DIFF
--- a/client/src/lib/components/Icons/Icon.tsx
+++ b/client/src/lib/components/Icons/Icon.tsx
@@ -7,7 +7,12 @@ const SvgIcon = styled(Svg)({
 });
 
 const Icon: React.FC<SvgProps> = ({children, style}) => (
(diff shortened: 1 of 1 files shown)


transient-jonas (approved):
Seems like `fill="transparent"` is the proper "setting" for all our icons, then looks good 👍

gewfy:
> Seems like `fill="transparent"` is the proper "setting" for all our icons, then looks good 👍

This is basically only for the root component. Also I believe this is the default for SVG in general.

PR #1737, opened 2023-02-06
Backend support for text sharings

swemail:


shared/src/types/Post.ts
```diff
@@ -0,0 +1,24 @@
+import type {Timestamp} from 'firebase-admin/firestore';
+import {LANGUAGE_TAG} from '../constants/i18n';
+import {UserProfile} from './User';
+
+export type PostFields = {
+  id: string;
+  exerciseId: string;
+  userId?: string;
```
gewfy, line 8: I guess this is only used for the anonymity switch and not to keep track of your own sharings?

functions/src/api/posts/index.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+        {
+          text: 'some new text',
+        },
+      );
+      expect(response.status).toBe(200);
+    });
+  });
+
+  describe('delete', () => {
+    it('should update post with public flag', async () => {
```
gewfy, line 140: Not correct description, right?

gewfy (commented):
Hmm I'm not really following this `public` concept. Can you elaborate?

shared/src/types/Post.ts
```diff
@@ -0,0 +1,24 @@
...
+  language: LANGUAGE_TAG;
+  public: boolean;
```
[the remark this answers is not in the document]
swemail: I think I will remove that, we will not send private messages to the server. If a user switches from public to private we delete and if add again if switching back

shared/src/types/Post.ts
swemail: That was the idea yes but I think I didn't follow the thought through 😄

shared/src/types/Post.ts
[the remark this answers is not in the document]
gewfy: 👌

firestore.indexes.json
```diff
@@ -63,6 +63,28 @@
...
+          "order": "ASCENDING"
+        },
+        {
+          "fieldPath": "approved",
+          "order": "ASCENDING"
+        },
+        {
+          "fieldPath": "public",
+          "order": "ASCENDING"
+        },
```
gewfy, line 82: This should be removed, right?

gewfy (approved):

firestore.indexes.json
swemail: Yes 😄

PR #1760, opened 2023-02-08
Separate live and async naming

gewfy:
This renames `Session` type to `LiveSession`. So we have a clear distinction between `LiveSession` and `AsyncSession` types.


transient-jonas (approved):
Nice 🎖️

PR #1769, opened 2023-02-09
Refactor async session modal into completed session modal

transient-jonas:
Make Modal reusable among completed async sessions and live completed sessions.


swemail (approved):
Great!

PR #1774, opened 2023-02-10
Remove separator from journey list

transient-jonas:
| Before | After |
|-|-|
| Screenshot 2023-02-10 at 09 04 35 |  |


gewfy (approved):
Phew that was a bunch. Could you please split your PRs into smaller chunks going forward...

PR #1825, opened 2023-02-13
Adjust journey scroll offset

transient-jonas:
Tested on Android only...

Could @gewfy or @swemail test it on ios as well?

Should scroll to such to a position like this:


swemail (approved):
Works on my machine 😄

transient-jonas:
THANK GOD SARAVÁ JESUS CHRIST OXALÁ AMEN 🙏 🥳 🌟

PR #1830, opened 2023-02-14
Remove unecessary completed session sorting on Joruney

transient-jonas:
I added this sorting not long ago thinking it was necessary but it was actually a brain fart 😂 
the completed sessions will already naturally be in order of completion (oldest first, closer to "now" in the bottom)

Exactly as by design, so the sorting isn't needed at all at the moment 😬


gewfy (approved):
You and your code deletions! Remember, it's not a competition 😄

transient-jonas:
😂

PR #1874, opened 2023-02-17
Move completedSessions to userEvents

swemail:


client/src/lib/sessions/hooks/useCompletedSessions.ts
```diff
@@ -1,13 +1,17 @@
...
-  const userState = useCurrentUserState();
+  const {completedSessionEvents} = useUserEvents();
 
   const completedSessions = useMemo(
-    () => userState?.completedSessions ?? [],
-    [userState],
+    () =>
+      completedSessionEvents.sort(
+        (a, b) =>
+          new Date(a.timestamp).getTime() - new Date(b.timestamp).getTime(),
```
transient-jonas, line 12: Stupid question: does this mean that a being more recent will go to the top?

Cause on the journey use case, we would display the most recent ones in the end of the list's section

client/src/lib/user/state/migration/v3.ts
```diff
@@ -0,0 +1,133 @@
...
+  userEvents?: Array<V3UserEvent>;
+};
+
+export type V3State = {
+  userState: {[key: string]: V3UserState};
+};
+
+const migrateCompletedSessionsToEvents = (
+  completedSessions: V3CompletedSession[],
+): CurrentUserEvent[] => {
```
transient-jonas, line 83: wouldn't it be more accurate to say `: CompletedSessionEvent[] =>`?

client/src/lib/user/state/migration/v3.ts
```diff
@@ -0,0 +1,133 @@
...
+  const sessions = uniqBy(s => s.id, completedSessions);
+  return sessions.map(s => ({
+    type: 'completedSession',
+    payload: omit(['completedAt'], s),
+    timestamp: new Date(s.completedAt),
+  })) as unknown as CurrentUserEvent[];
```
transient-jonas, line 89: Strange, the unknown thing again?
Is it because of ramda usage? cause I think I could do that without needing `as` at all, it was something like this:
```typescript
({ completedAt , ...s }) => ({
  type: 'completedSession',
  payload: s, 
  timestamp: new Date(completedAt)
})`

Comment (transient-jonas) [reply client/src/lib/user/state/migration/v3.ts:89]: Aah I get it, it is because of `type: V3SessionType;` it seems

Comment (transient-jonas) [reply client/src/lib/user/state/migration/v3.ts:89]: Does that kinda defeats the purpose of using the version types?

**Review by transient-jonas — APPROVED** [review]
A few comments to take into consideration but everything looks solid code-wise 👍 
Not sure about the completed sessions order, thing, I guess we should check that at least, before merging 🚀

File: client/src/lib/sessions/hooks/useCompletedSessions.ts (hunk 1 above)
Comment (swemail) [reply client/src/lib/sessions/hooks/useCompletedSessions.ts:12]: No, the opposite, latest last

File: client/src/lib/user/state/migration/v3.ts (hunk 3 above)
Comment (swemail) [reply client/src/lib/user/state/migration/v3.ts]: This is a special case with the unions, same issue in the user state file where I added a `getTypedEvent` to cast

Comment (transient-jonas) [reply client/src/lib/user/state/migration/v3.ts]: A way to make the `as` unnecessary is to update the `V3CompletedSession` fields to:

```typescript
type: SessionType.private | SessionType.public;
mode: SessionMode.async | SessionMode.live;
```

transient-jonas: Hmmm... not sure if I understand, on the `getTypedEvent` is different no?
Like, you're not returning `as unkown as ...` there, the type check actually occurs and still works...

PR #1877, opened 2023-02-17
Improve exercise lookup performance in sessions

gewfy:


transient-jonas:
Session after this change:

transient-jonas (approved):
🎉

PR #1884, opened 2023-02-20
Add Re-auth modal for user deletion

transient-jonas:
[Notion Task](https://www.notion.so/29k/Early-Access-2794500652b34c64b0aff0dbbc53e0ab?pvs=4#ddb01980a58642e899304189ce46fa15)

### Description of the Change

As a user when deleting your data, you should re-authenticate to both confirm your choice (UX) and so the session hasn't "expired" (tech).

Unless you're an anonymous user, then you don't need to authenticate.

#### Screenshots

https://user-images.githubusercontent.com/7523828/220126570-746f0259-0e5b-4216-a04c-41497cf757d6.mov

### DIFF
--- a/client/src/routes/DeleteUserModal/DeleteUserModal.tsx
+++ b/client/src/routes/DeleteUserModal/DeleteUserModal.tsx
@@ -36,9 +36,9 @@ const DeleteUserModal = () => {
   const [email, setEmail] = useState('');
   const [password, setPassword] = useState('');
   const [error, setError] = useState('');
-  const {deleteUser} = useDeleteUser();
+  const {deleteUser, isDeletingUser} = useDeleteUser();
 
-  const signIn = useCallback(async () => {
+  const onConfirm = useCallback(async () => {
     try {
       setIsSigningIn(true);
       await auth().signInWithEmailAndPassword(email, password);
@@ -63,7 +63,7 @@ const DeleteUserModal = () => {
             autoCapitalize="none"
             autoComplete="email"
             autoCorrect={false}
-            onSubmitEditing={signIn}
+            onSubmitEditing={onConfirm}
             placeholder={t('email')}
             onChangeText={setEmail}
             defaultValue={email}
@@ -74,7 +74,7 @@ const DeleteUserModal = () => {
             autoCapitalize="none"
             autoComplete="password"
             autoCorrect={false}
-            onSubmitEditing={signIn}
+            onSubmitEditing={onConfirm}
             placeholder={t('password')}
             onChangeText={setPassword}
           />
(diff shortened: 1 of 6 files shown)


client/src/routes/DeleteUserModal/DeleteUserModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+            <Error>{t(`errors.${error}`)}</Error>
+            <Spacer16 />
+          </>
+        )}
+        <StyledButton
+          variant="primary"
+          disabled={isSigningIn || !email || !password}
+          loading={isSigningIn}
+          onPress={signIn}>
+          {t('signIn')}
```
transient-jonas, line 94: need to update those keys and namings

client/src/routes/SignInModal/SignInModal.tsx
```diff
@@ -38,9 +38,6 @@ const SignInModal = () => {
   const signIn = useCallback(async () => {
     try {
       setIsSigningIn(true);
-      if (auth().currentUser) {
-        await auth().signOut();
-      }
```
gewfy, line 43: I guess you don't need to sign out before signing in? Or where there other reasons to remove this?

client/src/routes/SignInModal/SignInModal.tsx
```diff
@@ -87,7 +84,7 @@ const SignInModal = () => {
         )}
         <StyledButton
           variant="primary"
-          disabled={isSigningIn}
+          disabled={isSigningIn || !email || !password}
```
gewfy, line 87: Nice I beleive this would close #1885

client/src/routes/DeleteUserModal/DeleteUserModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+            textContentType="emailAddress"
+            keyboardType="email-address"
+            autoCapitalize="none"
+            autoComplete="email"
+            autoCorrect={false}
+            onSubmitEditing={signIn}
+            placeholder={t('email')}
+            onChangeText={setEmail}
+            defaultValue={email}
+          />
```
gewfy, line 70: I think we should remove the email address field and only require password. Sort of weird that you can sign in as another user to delete it.

gewfy (approved):
Works great and implementation is 👌 But I think we should only require password :)

client/src/routes/DeleteUserModal/DeleteUserModal.tsx
transient-jonas: ooooh yes, nice catch, totally!

Will update 👍

client/src/routes/SignInModal/SignInModal.tsx
transient-jonas: Exactly! I was wondering about that, so I tested it and it works without signing out.
What I observed was that if you try to sing in and "it doesn't work" you'll have signed out and not signed in either.

PR #1934, opened 2023-02-21
Completed session flow and user profile caching

transient-jonas:
### Tasks
- [Completed session task](https://www.notion.so/29k/Early-Access-2794500652b34c64b0aff0dbbc53e0ab?pvs=4#c6567afdeb7840a4a73ebea347b9af18)
- [Host profile caching](https://www.notion.so/29k/Early-Access-2794500652b34c64b0aff0dbbc53e0ab?pvs=4#8fe5450d833d4884a55db75082ba2679)

### Description of the Change

- Requests to `GET /user/:id` gets cached by the client for 30min (automatically through response header).
- The `useUserProfile` hook either fetches the profile or uses the current user state to get it (in case you're the host)
- All session modes/types follow the same flow when getting completed.


transient-jonas:
The session completed flow is not working as expected, continuing debugging it

client/src/routes/modals/SessionFeedbackModal/SessionFeedbackModal.tsx
```diff
@@ -96,7 +101,7 @@ const SessionFeedbackModal = () => {
   }, [snapToIndex, setAnswer]);
 
   const submit = useCallback(() => {
-    if (answer) {
```
transient-jonas, line 99: this is kinda funny btw, we never allowed answer to be thumbs down 😂

functions/src/api/user/index.ts
```diff
@@ -86,6 +86,7 @@ userRouter.put(
 userRouter.get('/:id', async ctx => {
   try {
     const userProfile = await getProfile(ctx.params.id);
+    ctx.set('Cache-Control', 'max-age=1800');
```
transient-jonas, line 89: 30min is enough right? or?

client/src/routes/modals/SessionFeedbackModal/SessionFeedbackModal.tsx
swemail: 😂

swemail (approved):
LGTM 🚀

PR #1935, opened 2023-02-22
Use string instead of date for event timestamp

swemail:
Since Date is serialized to string i async storage it is more accurate to type it as string. Also we get warnings when passing `CompletedSessionEvent` as route param before it as gone through a round trip to the storage.

Had to be a history revisionist in the migration scripts but It's actually more accurate for the same reason as mentioned above


gewfy:
Related to this issue? #1938

gewfy (approved):
👌

swemail:
> Related to this issue? #1938

Most definitely

PR #1936, opened 2023-02-22
Dev button to clear stored events and pinned sessions

swemail:
The journey tab fills up during development, added a dev button to clear user events and pinned sessions


gewfy:
Isn't delete all data button enough? :P

PR #1937, opened 2023-02-22
Update email field

transient-jonas:
Not quite the same spacing but I tried using the closest variable we have.

Screenshot 2023-02-22 at 08 55 40

Screenshot 2023-02-22 at 08 53 30

### DIFF
--- a/client/src/routes/DeleteUserModal/DeleteUserModal.tsx
+++ b/client/src/routes/DeleteUserModal/DeleteUserModal.tsx
@@ -25,7 +25,8 @@ const StyledButton = styled(Button)({
 });
 
 const Row = styled.View({
-  padding: SPACINGS.SIXTEEN,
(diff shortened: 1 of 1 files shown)


client/src/routes/DeleteUserModal/DeleteUserModal.tsx
```diff
@@ -25,7 +25,8 @@ const StyledButton = styled(Button)({
 });
 
 const Row = styled.View({
-  padding: SPACINGS.SIXTEEN,
+  paddingLeft: SPACINGS.SIXTEEN,
+  padding: SPACINGS.TWELVE,
```
transient-jonas, line 29: oops, need to update this to bottom, right, etc

gewfy (approved):
😄  Thanks for the very descriptive screenshots 🤗

PR #1944, opened 2023-02-24
Async sharing adjustments

swemail:
- [x] Load posts while on previous slide to make it done by the time you arrive there
- [x] Scroll to add sharing button while on previous slide
- [x] Sort your prev sharings as latest first
- [x] Close `CompletedSessionModal` when pressing "Do this exercise again" to not get back to it when `CreateSessionModal` is closed


transient-jonas (approved):
Noice!

PR #1980, opened 2023-02-28
Implement closingTime logic for sessions

transient-jonas:
Related to [this notion task](https://www.notion.so/29k/Early-Access-2794500652b34c64b0aff0dbbc53e0ab?pvs=4#1228dd30e0614599886bed243445127f)

### Description of the Change

This will simplify the filtering for `closed` sessions altogether, 
if we don't do this we would need to make 2 queries.

https://github.com/29ki/29k/blob/87c78584f59de2ed04cdff62d46a14a68ed3308d/functions/src/models/session.ts#L87-L97

If we would try to filter through a separate field we'd need to define it a default value as far ahead of time so sessions that didn't start wouldn't be filtered out, or do two queries: started sessions, not started sessions and combine, so why not refactor the way we filter for `startTime` + 30min and do it with one field only, `closingTime`.

Let me know your thoughts!

### DIFF
--- a/functions/src/controllers/sessions.ts
+++ b/functions/src/controllers/sessions.ts
@@ -206,6 +206,12 @@ export const updateSessionState = async (
     sessionModel.updateSession(sessionId, {ended: true});
   }
 
+  if (data.started) {
+    sessionModel.updateSession(sessionId, {
+      closingTime: dayjs().add(5, 'minutes').toString(),
+    });
+  }
+
   await sessionModel.updateSessionState(sessionId, removeEmpty(data));
   const updatedState = await sessionModel.getSessionStateById(sessionId);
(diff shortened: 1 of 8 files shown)


functions/src/controllers/sessions.ts
```diff
@@ -206,6 +206,12 @@ export const updateSessionState = async (
     sessionModel.updateSession(sessionId, {ended: true});
   }
 
+  if (data.started) {
+    sessionModel.updateSession(sessionId, {
+      closingTime: dayjs().add(5, 'minutes').toString(),
```
transient-jonas, line 211: should add tests for this

swemail (approved):
Great, looks good!

transient-jonas:
Adding a migration script and firestore indexes soon

PR #1985, opened 2023-03-01
Ongoing sessions

transient-jonas:
Issue: [Notion Task](https://www.notion.so/29k/Early-Access-2794500652b34c64b0aff0dbbc53e0ab?pvs=4#1228dd30e0614599886bed243445127f)


content/src/ui/Screen.Home.json
```diff
@@ -5,7 +5,8 @@
     "sections": {
       "hostedBy": "Your sessions",
       "interested": "Interested",
-      "comming": "Upcoming sessions"
+      "comming": "Upcoming sessions",
+      "ongoing": "Ongoing sessions"
```
transient-jonas, line 9: need to remove these 😄

client/src/routes/screens/LiveSession/ChangingRoom.tsx
```diff
@@ -191,12 +191,19 @@ const ChangingRoom = () => {
     session,
   ]);
 
-  const joinPress = useCallback(() => {
+  const joinPress = useCallback(async () => {
+    joinSession(session.inviteCode);
```
gewfy, line 195: Ah interesting approach. In the discussion we talked about doing this in the `preJoin` https://github.com/29ki/29k/blob/main/client/src/routes/screens/LiveSession/ChangingRoom.tsx#L149

But I guess this is better because it's when the user actually is committed. But FYI for all of us @swemail

gewfy (approved):
Nice nice nice! Maybe merge this after todays session, so we can test it out in staging?

PR #1992, opened 2023-03-02
Change default async session type

gewfy:
I managed to flag all async sessions as `private` when they are in fact `public`. This flags them appropriately and migrates all `completedSession` events.


swemail (approved):
How awesome then that we can just migrate 👌

PR #2038, opened 2023-03-07
Query posts by sharing id

swemail:
Fixes bug here we didn't differentiate between sharings when session has several sharings


gewfy (approved):
Nice! Looked like a smooth addition

PR #2043, opened 2023-03-07
Fix re-auth on updating profile email/password

transient-jonas:
Issue: [Notion Task 🦾 ](https://www.notion.so/29k/Early-Access-2794500652b34c64b0aff0dbbc53e0ab?pvs=4#4dfb97222a3a444493f5c3789faf7f34)

### Description of the Change

Changing email and/or password requires a recent login (like deleting account) 

#### Screenshots
|  |  |  |
| ------ | ------ | ------ |
|  |   |  |


swemail (approved):

gewfy (approved):
Really nice 👌 Just to double check, if you change your display name it doesn't require password? Only some failing unit testst :)

client/src/routes/modals/ProfileSettingsModal/ProfileSettingsModal.tsx
```diff
@@ -142,6 +150,18 @@ const ProfileSettingsModal = () => {
               onChangeText={setEmail}
               defaultValue={email}
             />
+            {user && !user?.isAnonymous && (
+              <BottomSheetActionTextInput
+                textContentType="currentPassword"
+                secureTextEntry
+                autoCapitalize="none"
+                autoComplete="password-current"
```
gewfy, line 158: Oh I've done this too and we got errors in Sentry about it. If I remember correctly this is not available until RN 0.71. I think it needs to be `password` only https://reactnative.dev/docs/0.69/textinput#autocomplete-android

client/src/routes/modals/ProfileSettingsModal/ProfileSettingsModal.tsx
```diff
@@ -142,6 +150,18 @@ const ProfileSettingsModal = () => {
               onChangeText={setEmail}
               defaultValue={email}
             />
+            {user && !user?.isAnonymous && (
+              <BottomSheetActionTextInput
+                textContentType="currentPassword"
```
gewfy, line 155: And this should be `password` I beleive

client/src/routes/modals/ProfileSettingsModal/ProfileSettingsModal.tsx
transient-jonas: aha, so I don't define it for the "newPassword" field I guess?

client/src/routes/modals/ProfileSettingsModal/ProfileSettingsModal.tsx
transient-jonas: updated it 👍

client/src/routes/modals/ProfileSettingsModal/ProfileSettingsModal.tsx
transient-jonas: Updated 👍 
I left the new password filed without the autocomplete field

PR #2089, opened 2023-03-14
Safety toolkit

transient-jonas:
Click the `Preview` tab and select a PR template:

- [Default template](?quick_pull=1&template=default.md)
- [Native release checklist](?quick_pull=1&template=native_release_checklist.md&title=🚢+Release+2.x.x&labels=native+release)


gewfy (approved):
Looks solid!!
Two things:
1. I think you could have split this up in multiple PRs, like one with only the back-end
2. Not possible to edit the emails in CMS yet? Would be nice to add. Not in this PR though.
3. Has the `SENDGRID_API_KEY` been added to workflows and GitHub secrets. Thinking about where to add som guidelines on what needs to be done every time we add env variables.

transient-jonas:
Yes, the variable got added to the workflow and secrets

PR #2211, opened 2023-04-03
Non authorized api routes

gewfy:
For the upcoming onboarding we want to create an anonymous user as late as possible. Today `killSwitch` requires a user and an authorization token. This adds the possibility to have non authorized api routes.

- [x] Add `preAuthRouter` support to api for routes that doesn't require authorization
- [x] Add support for non authorized api requests in `apiClient` with `authorize = false`
- [x] Change `killSwitch` to be a non authorized route


swemail (approved):
Nice!

PR #2214, opened 2023-04-03
Upcoming public sessions

gewfy:
In preparation for the onboarding. This adds a non authorised route for fetching the closest upcoming, public sessions `/api/onboarding/sessions`


swemail (approved):
👌

PR #2241, opened 2023-04-04
Rating

swemail:
Adds rating modal 
<image src="https://user-images.githubusercontent.com/1139207/230013309-3321e93f-51a5-40bb-a468-59885dfd76d9.png" width="200" />


client/ios/twentyninek.xcodeproj/project.pbxproj
gewfy, line 1: These changes doesn’t look right. Or?

.github/workflows/build-ios.yml
gewfy, line 1: Since these are used in the js bundle I believe you need to add both to iOS, android and codepush

client/src/lib/appState/hooks/useRating.ts
gewfy, line 1: I would put these hooks in a separate rating lib. It’s not really appstate.

client/src/routes/modals/CompletedSessionModal/CompletedSessionModal.tsx
```diff
@@ -86,9 +87,16 @@ const CompletedSessionModal = () => {
...
   const user = useUser();
+  const askForRating = useRating();
 
   const sessionTime = useMemo(() => dayjs(timestamp), [timestamp]);
 
+  useEffect(() => {
+    if (askForRating(completedSessionEvent.payload.id)) {
+      navigate('RatingModal');
+    }
+  }, [askForRating, navigate, completedSessionEvent.payload.id]);
```
gewfy, line 98: Why are we asking about this reactively here instead of when submitting feedback?

client/ios/twentyninek.xcodeproj/project.pbxproj
swemail: No, sorry, I have to change to that to build on my phone and I need to change back before push, forgot, will revert

gewfy (approved):

PR #2244, opened 2023-04-04
Include session type and mode in feedback

swemail:
So that we can know in what context the feedback is given

### DIFF
--- a/client/src/lib/session/hooks/useLeaveSession.ts
+++ b/client/src/lib/session/hooks/useLeaveSession.ts
@@ -13,14 +13,20 @@ import {
 import useSessions from '../../../lib/sessions/hooks/useSessions';
 import useLiveSessionMetricEvents from './useLiveSessionMetricEvents';
 import useIsSessionHost from './useIsSessionHost';
-import {SessionMode} from '../../../../../shared/src/types/Session';
+import {
+  SessionMode,
+  SessionType,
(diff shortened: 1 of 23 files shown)


client/src/lib/session/hooks/useLeaveSession.ts
```diff
@@ -13,14 +13,20 @@ import {
...
 
 type ScreenNavigationProps = NativeStackNavigationProp<
   TabNavigatorProps & ModalStackProps
 >;
 
-const useLeaveSession = (sessionMode: SessionMode) => {
+const useLeaveSession = (
+  sessionMode: SessionMode,
+  sessionType: SessionType,
+) => {
```
gewfy, line 29: How about sending the full session object here instead? Not a biggie, just feels a bit cleaner so I leave it up to you.

gewfy (approved):
Solid as a rock! Minor nit pick only. 
I'm pondering how we could make this type of flow of passing data a bit more cleaner. Something feels clunky with how we setup the sate, pass data into components and hooks and then teardown of state in relation to data that is needed after the session.
No answers or changes needed. Just thinking out loud...

client/src/lib/session/hooks/useLeaveSession.ts
swemail: Had that thought, so now there are two similar thoughts on that topic 😄  That's enough for me to change it

PR #2255, opened 2023-04-06
Add email related content to CMS

jmpas:
Screenshot 2023-04-06 at 17 11 09
Screenshot 2023-04-06 at 17 11 04
Screenshot 2023-04-06 at 17 10 59

### DIFF
--- a/cms/src/lib/i18n.ts
+++ b/cms/src/lib/i18n.ts
@@ -93,7 +93,7 @@ export const generateFieldFromI18n = ([key, value]: [
 export const generateFilesCollectionFromi18nFiles = (
   folderName: string,
   label: string,
-  i18nResources: Resource,
+  i18nResources: ResourceLanguage,
   exclude: Array<string>,
 ): CmsCollection => {
   const filteredResources = Object.entries(i18nResources).filter(


cms/src/collections/collections.ts
```diff
@@ -110,9 +110,22 @@ export const other: CmsCollection = {
...
+
 export const ui: CmsCollection = generateFilesCollectionFromi18nFiles(
   'ui',
   '📱 UI',
   content.i18n,
-  [exercises.name, tags.name, collections.name],
+  resouce =>
+    ![exercises.name, tags.name, collections.name].find(
+      e => e === resouce[0],
+    ) && !resouce[0].includes('email.'),
```
gewfy, line 130: I'm trying to understand this but can't wrap my head around. Maybe it has become more complex than it has to be? Could we do a round of simplifying this so we don't need to filter out stuff here?

transient-jonas: Would it be simpler/more clear if the callback would receive `resourceFileName` and `resourceContent` ?
Those are the `resource[0]`  and `resource[1]`.

or do you have any ideas on how it could be simpler in a even higher level? (another approach maybe)

transient-jonas: Exclude callback type:
https://github.com/29ki/29k/blob/b9ee222955f8bc16f8a85ef6b157df04cc32093c/cms/src/lib/i18n.ts#L97

gewfy (approved):

PR #2256, opened 2023-04-06
Host sheets

swemail:
Show host info modal. Accessible from sessions cards and hosts modal.

Public host flow is now changed:
* Submitting verification code now creates an entry in the `users` collection with `role: "publicHost"`
* Firestore trigger listening to an update to the `role` field and updates user claim with that role
* Enables us to also remove the public host role if needed by just removing the role on the user record

This is so that we can list public hosts from the users collection and access the profile of the user to show in the host info modal

<image src="https://user-images.githubusercontent.com/1139207/231984224-2bc25e5b-aff5-4a83-9ed6-6ac14fe51239.png" width="200" /><image src="https://user-images.githubusercontent.com/1139207/231984864-afedf2f7-1031-4d02-88c3-a16ce5c588c2.png" width="200" />

Also add the possibility to add a description on your profile

<image src="https://user-images.githubusercontent.com/1139207/231996616-3a50657b-7477-429a-a7b1-44cbdf2c8825.png" width="200" />


client/src/lib/components/Cards/Card.tsx
```diff
@@ -111,7 +113,13 @@ export const Card: React.FC<CardProps> = ({
           <Header>
             {title && <Display20 numberOfLines={2}>{title}</Display20>}
             <Spacer4 />
-            <Byline pictureURL={hostPictureURL} name={hostName} />
+            {onHostPress ? (
```
transient-jonas, line 116: Another way of doing this is by defining an empty callback as default for the onHostPress, and maybe `disabled={!onHostPress}` if we wanna make it explicit I guess.

But the current approach is also legit, just pointing another possible way of doing it 😁

transient-jonas (approved):
Super cool man! good work 🤘

client/src/lib/components/Cards/Card.tsx
swemail: Ah, right, the disabled would mitigate that I guess

PR #2299, opened 2023-04-12
Local video

swemail:
Adds possibility to play local assets added with `react-native-asset` Specify as filename with extension i.e. `my_video_asset.mp4`


gewfy (approved):
Boooootiful!

Issue #2317, opened 2023-04-17
Incoming calls interrupts Intro portal video and is never resumed


swemail:
### Description

When being in the intro portal and getting a call, the video and audio is paused and never resumed

### Steps to Reproduce

1. Create a live session
2. Enter the intro portal
3. Call yor phone

**Expected behavior:**

When call has ended or was declined, the video and audio should resume

**Actual behavior:**

Both video and audio is paused and you are stuck in the intro portal

PR #2318, opened 2023-04-17
Handle call interruptions

swemail:
Fixes issues for iOS. Android behaves good enough. The sound is interrupted on Android if you answer a call but I assume most will decline calls if they are in a session.
Solves #2317


gewfy (approved):
Nice! I understand 20% of it :D

PR #2320, opened 2023-04-17
Remove pt as fully supported language

swemail:


gewfy (approved):
Not enough changes! More change! :D

PR #2355, opened 2023-04-18
Count hosted session

swemail:
Update count for the host when session reaches completed state.

<image src="https://user-images.githubusercontent.com/1139207/232753615-b871b5ea-4e71-422d-9ae9-edabf71560dc.png" width="200" />

### DIFF
--- a/functions/src/models/user.ts
+++ b/functions/src/models/user.ts
@@ -18,7 +18,7 @@ export const updateUser = async (id: string, userData: Partial<UserData>) => {
     .set({...userData, updatedAt: now}, {merge: true});
 };
 
-export const updateHostedCount = async (
+export const incrementHostedCount = async (
   id: string,
   countProperty: keyof HostedCount,
 ) => {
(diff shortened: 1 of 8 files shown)


functions/src/models/user.ts
```diff
@@ -18,6 +18,26 @@ export const updateUser = async (id: string, userData: Partial<UserData>) => {
     .set({...userData, updatedAt: now}, {merge: true});
 };
 
+export const updateHostedCount = async (
```
gewfy, line 21: Could we call it `incrementHostedCount`?

gewfy (approved):
Nice! I'd like some response filtering, especially for the user end-point. But let's discuss that on heap tomorrow.

swemail:
> Nice! I'd like some response filtering, especially for the user end-point. But let's discuss that on heap tomorrow.

Yea I thought about it but since we where about to talk about it I made it waited for now

PR #2369, opened 2023-04-19
Add response filters as middleware

swemail:


gewfy:
image

gewfy:
giphy (1)

client/src/lib/session/hooks/useSubscribeToSession.ts
```diff
@@ -22,12 +20,7 @@ const useSubscribeToSession = (sessionId: LiveSession['id']) => {
           if (!snapshot.exists) {
             onSnapshot(undefined);
           }
-
-          const sessionState = getSessionState(
-            getData<SessionStateData>(snapshot),
-          );
-
-          onSnapshot(sessionState);
+          onSnapshot(SessionStateSchema.validateSync(snapshot.data()));
```
gewfy, line 23: Why do we need to validate on this end?

functions/src/api/lib/validation.ts
```diff
@@ -0,0 +1,134 @@
...
+        throw error;
+      }
+    }
+  };
+
+export interface ResponseContext extends Koa.Context {
+  state: ValidatedState;
+}
+
+export const assertValidatedResponse =
```
gewfy, line 121: Shouldn't this assert request validation too? Even if the request might be empty

functions/src/api/lib/validation.ts
```diff
@@ -0,0 +1,134 @@
...
+          options.body,
+        );
+      }
+
+      if (validator.query) {
+        ctx.state.query = await validator.query.validate(
+          ctx.request.query,
+          options.query,
+        );
+      }
```
gewfy, line 89: We should add validation `params` and maybe `headers` too?

functions/src/api/lib/validation.ts
```diff
@@ -0,0 +1,134 @@
...
+    } catch (error) {
+      if (error instanceof yup.ValidationError) {
+        throw new RequestValidationError('Request validation failed', error);
```
gewfy, line 92: Curious, does this give us more explicit errors in dev mode now? :)

functions/src/api/lib/validation.ts
```diff
@@ -0,0 +1,134 @@
...
+    TQuery extends yup.Schema,
+    TResponse extends yup.Schema,
+  >(
+    validator: Partial<Validator<TBody, TQuery, TResponse>>,
+    options: ValidatorOptions = defaultOptions,
+  ): Koa.Middleware<ValidatedState<TBody, TQuery, TResponse>> =>
+  async (ctx: Koa.Context, next: Koa.Next) => {
+    try {
+      if (validator.body) {
+        ctx.state.body = await validator.body.validate(
```
gewfy, line 78: No possibility to update `ctx.body` and `ctx.query` directly here, since it's the default place to pick data from?

shared/src/modelUtils/session.ts
gewfy, line 1: Nice that there's no need for this anymore 👏

shared/src/types/Post.ts
```diff
@@ -1,28 +1,43 @@
...
+]).concat(
+  yup.object({
+    anonymous: yup.boolean().default(true),
+    language: yup
+      .mixed<LANGUAGE_TAG>()
+      .oneOf(LANGUAGE_TAGS)
+      .default(DEFAULT_LANGUAGE_TAG),
+  }),
+);
+export type CreatePostData = yup.InferType<typeof CreatePostSchema>;
```
gewfy, line 43: Nit pick, but maybe these should be called `CreatePostDataType` or `CreatePostType` just to not confuse them with schemas. Only if you agree.

shared/src/types/Post.ts
gewfy, line 1: How about renaming `shared/src/types` to `schemas`?

gewfy (commented):
Wow what a beautiful beast! 😍

client/src/lib/session/hooks/useSubscribeToSession.ts
swemail: Since this is directly from the database we get the transformation with this

functions/src/api/lib/validation.ts
swemail: I had that as a first thought but couldn't get it to run in the correct stage. Then I thought that request validation is a bit more explicit since we want to get the types in the handlers so I removed the assertion

functions/src/api/lib/validation.ts
swemail: No, the Koa types didn't seem to approve of that, will check once more

gewfy: Last time I checked body is an object. But query is some kind of prototype with helpers.

functions/src/api/lib/validation.ts
gewfy: Got it. Then I think we should either manipulate the incoming data (and not `ctx.state`) or assert request schemas. Otherwise I see us create endpoints without request validation and/or not picking data from `ctx.state`.

functions/src/api/lib/validation.ts
swemail: guess this is the reason koa-yup-validate is using the state object for this

functions/src/api/lib/validation.ts
```diff
@@ -0,0 +1,152 @@
...
+const defaultOptions: ValidatorOptions = {
+  body: {
+    stripUnknown: true,
+  },
+  query: {
+    stripUnknown: true,
+  },
+  response: {
+    stripUnknown: true,
+  },
```
gewfy, line 73: Did you give up on `params` validation? I think we should have that.

gewfy (approved):

PR #2485, opened 2023-04-27
Show video while navigating to session

gewfy:


swemail (approved):
Bonafide swift developer 👌

PR #2578, opened 2023-05-05
Restart video reaching end issue

swemail:
[Video reset issue](https://www.notion.so/29k/Early-Access-2794500652b34c64b0aff0dbbc53e0ab?pvs=4#2c9ff74097004081a14678b0d69c0548)


client/src/lib/session/components/Slides/Slides/Blocks/Video.tsx
```diff
@@ -63,6 +63,8 @@ const Video: React.FC<VideoProps> = ({
 
       // Reset onEndRef when playing
       if (playing) {
+        console.log('reset');
```
transient-jonas, line 66: debug code?

transient-jonas (approved):

client/src/lib/session/components/Slides/Slides/Blocks/Video.tsx
```diff
@@ -106,9 +108,10 @@ const Video: React.FC<VideoProps> = ({
     // Check that the progressRef is not set back to zero to trigger a reset
```
gewfy, line 108: Is this comment still relevant?

client/src/lib/session/components/Slides/Slides/Blocks/Video.tsx
```diff
@@ -106,9 +108,10 @@ const Video: React.FC<VideoProps> = ({
...
     if (!autoPlayLoop && !onEndRef.current) {
       onEndRef.current = true;
+      seek(0);
```
gewfy, line 111: I don't think this is what we want. Before the video just stopped when coming to an end. And then if you press the rewind button it started playing again as the playing state was already playing.

PR #2585, opened 2023-05-08
Fix profile pic journey

transient-jonas:
Ref: https://29k.slack.com/archives/CH1RR6XCL/p1683287555045649

| no profile pic | with profile pic |
|-|-|
|Simulator Screen Shot - iPhone 14 Plus - 2023-05-08 at 14 54 47 |  Simulator Screen Shot - iPhone 14 Plus - 2023-05-08 at 14 54 37 |


transient-jonas:
Wrong size, will push a fix

transient-jonas:
Note, no need to overthink the situation where host removes their pic, that shouldn't be a common use case

swemail (approved):
Nice, not sure why it didn't work with just Image

PR #2615, opened 2023-05-09
Use setUserData to update if in portal

swemail:
join meeting overwrote the user data already set, removing userName

### DIFF
--- a/client/src/routes/screens/LiveSession/ChangingRoom.tsx
+++ b/client/src/routes/screens/LiveSession/ChangingRoom.tsx
@@ -116,8 +116,14 @@ const ChangingRoom = () => {
         LiveSessionStackProps & TabNavigatorProps & ModalStackProps
       >
     >();
-  const {toggleAudio, toggleVideo, setUserName, joinMeeting, preJoinMeeting} =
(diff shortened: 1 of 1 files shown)


client/src/routes/screens/LiveSession/ChangingRoom.tsx
```diff
@@ -177,17 +183,16 @@ const ChangingRoom = () => {
     } else {
       await joinMeeting({
         subscribeToTracksAutomatically: false,
-        userData: {
-          inPortal: true,
-        },
       });
+      setUserData({inPortal: true});
```
gewfy, line 187: My worry with this approach is that the data is not directly on joining, but hopefully asyncrounous short there after. Aaaaand at the same time we're going to remove this completely soon (I hope)

gewfy (approved):

client/src/routes/screens/LiveSession/ChangingRoom.tsx
swemail: Yea, know, but it's going to be a very short while before it's visible users are waiting in the protal

PR #2662, opened 2023-05-11
Temp fix to not get the timer started before press play

swemail:
There is a race in the `LottiePlayer` preventing it to pause the player in the useEffect. This is a temp fix since that will take some investigation


gewfy (approved):
Interesting! Could you add a todo in notion so we don't forget to follow-up?

PR #2694, opened 2023-05-15
Add link to collections and exercises

swemail:
Some type name updates also got included by running the type generation from the cms


gewfy (approved):
I changed the link to `duplicate` between translations for now. Also added the already created links for the available collections.

PR #2749, opened 2023-05-16
Check in client that session is still open

swemail:
Found this the least intrusive way. The user still have to wait for the pre-join round trip, so might as well check here

### DIFF
--- a/client/src/lib/session/hooks/useIsAllowedToJoin.spec.ts
+++ b/client/src/lib/session/hooks/useIsAllowedToJoin.spec.ts
@@ -1,0 +1,144 @@
+import {useNavigation} from '@react-navigation/native';
+import {renderHook} from '@testing-library/react-hooks';
+import {FirebaseAuthTypes} from '@react-native-firebase/auth';
+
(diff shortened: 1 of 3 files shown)


functions/src/controllers/sessions.ts
```diff
@@ -105,6 +105,10 @@ export const getSession = async (
     throw new RequestError(ValidateSessionError.userNotFound);
   }
 
+  if (!isUserAllowedToJoin(session, userId)) {
```
gewfy, line 108: This check and the one above does the same thing checking if the user is allowed to join. Maybe we only need one here? 
Sometimes we do this user check in the controller, and sometimes in the firestore query. Maybe we should do it the same way always? - maybe no in this PR, but as a next improvement?

gewfy (approved):
Looks good! One smaller improvement :)

functions/src/controllers/sessions.ts
swemail: Yea, lump it together, user is in and it's private, or the time check. I'll release now and write a note

PR #2804, opened 2023-05-22
Refresh sessions intermittently

transient-jonas:
ref: https://www.notion.so/29k/Aware-app-e8b9a3d421cc499f94d0cc1a0f7f51d4?pvs=4#296a809e7a1443fc8026ae7ff274ce6e


client/src/routes/screens/Home/Home.tsx
```diff
@@ -217,6 +224,23 @@ const Home = () => {
     navigate('AboutOverlay');
   }, [navigate]);
 
+  const throttledRefresh = useMemo(
+    () => throttle(() => refreshPull(), 5 * 60000),
```
gewfy, line 228: How about only doing a `fetchSessions` and not `refreshPull` here since it will pull down and show that it's loading otherwise? Less intrusive imo

client/src/routes/screens/Home/Home.tsx
```diff
@@ -217,6 +224,23 @@ const Home = () => {
...
+  useEffect(() => {
+    if (isFocused) {
+      throttledRefresh();
+    }
+  }, [isFocused, throttledRefresh]);
+
+  useResumeFromBackgrounded(() => {
+    if (isFocused) {
+      throttledRefresh();
+    }
```
gewfy, line 241: Since you do the same check in these two hooks, maybe you could move it inside of the throttled function?

client/src/routes/screens/Home/Home.tsx
gewfy: Also shouldn't this be a useCallback? E.g.
```javascript
const throttledFetch = useCallback(throttle(fetchSessions), 5 * 60000), [fetchSessions])
```

gewfy (changes requested):

client/src/routes/screens/Home/Home.tsx
transient-jonas: done [6479f65](https://github.com/29ki/29k/pull/2804/commits/6479f6536378430f562f65a514d08eac4ecd623c)

client/src/routes/screens/Home/Home.tsx
transient-jonas: Nice! I like the less intrusive perspective, done [6479f65](https://github.com/29ki/29k/pull/2804/commits/6479f6536378430f562f65a514d08eac4ecd623c)

gewfy (approved):
Very nice 👌

PR #2810, opened 2023-05-23
Fix black status bar in editorial

swemail:
| Before | After |
|-|-|
|<image src="https://github.com/29ki/29k/assets/1139207/dcfb441c-e1a2-4443-9d6b-2b6fd2eece31" width="200" />   |<image src="https://github.com/29ki/29k/assets/1139207/0e712e04-0efc-4220-bba8-d8ecc0544e78" width="200" />  |


gewfy:
Yeah I know this is the quick fix... but I would like to find a more automatic way or the actual bug. Like in this case you forgot about the community editorial ;)

swemail:
> Yeah I know this is the quick fix... but I would like to find a more automatic way or the actual bug. Like in this case you forgot about the community editorial ;)

Right, then it's a bit harder. Think this is lower prio then

PR #2860, opened 2023-05-25
Animated buttons

swemail:
The bell button is a bit delayed but that is not an animation issue so we can fix that later. 

<video src="https://github.com/29ki/29k/assets/1139207/776c4a16-0daf-447d-bcf3-e84bdaf8f299" width="200" />


client/src/lib/components/Icons/AnimatedBell/AnimatedBell.tsx
```diff
@@ -0,0 +1,371 @@
...
+        },
+      ],
+      ip: 0,
+      op: 576,
+      st: 0,
+      bm: 0,
+    },
+  ],
+  markers: [],
+};
```
gewfy, line 348: lol

client/src/lib/components/Icons/AnimatedBell/AnimatedBell.tsx
```diff
@@ -0,0 +1,371 @@
...
+  AnimatedLottieView,
+  AnimatedBellProps
+>(({fill = COLORS.BLACK, ...props}, ref) => {
+  const colorFilters = useMemo(
+    () => [
+      {keypath: 'Layer 1', color: fill},
+      {keypath: 'Layer 2', color: fill},
+    ],
+    [fill],
+  );
```
gewfy, line 362: Wow! Neat feature 😍

client/src/lib/components/Icons/AnimatedBell/AnimatedBell.tsx
gewfy: How about having these as a separate json-file in the same folder? :P

client/src/lib/components/Buttons/IconButton/AnimatedIconButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+import {AnimatedIconType} from '../../Icons/AnimatedIcon';
+
+export type BaseIconButtonProps = BaseButtonProps & {
+  noBackground?: boolean;
+  fill?: string;
+};
+
+const StyledIconButton = styled(BaseButton)<BaseIconButtonProps>(props => ({
+  ...(props.noBackground ? {backgroundColor: 'transparent'} : {}),
+  width: props.small ? 36 : 44,
```
transient-jonas, line 15: Do we usually use the `SPACINGS` for height and width too? if so we could use them here I think

swemail: Just coped it from the other buttons but true that

client/src/lib/components/Buttons/IconButton/AnimatedIconButton.tsx
```diff
@@ -0,0 +1,88 @@
...
+export type IconButtonProps = Omit<BaseIconButtonProps, 'variant'> & {
+  AnimatedIcon: AnimatedIconType;
+  preVariant: BaseButtonProps['variant'];
+  postVariant: BaseButtonProps['variant'];
+};
+
+const AnimatedIconButton: React.FC<IconButtonProps> = ({
+  onPress,
+  preVariant,
+  postVariant,
```
gewfy, line 30: Hmm do we have to have a pre and post here? Can't we just react to state changes in an effect on certain props? E.g always animate when variant changes? Now it's like the logic is inside of this component. 

... but I might be missing aspects of the problem.

client/src/lib/components/Icons/AnimatedBell/AnimatedBell.tsx
```diff
@@ -0,0 +1,371 @@
...
+      st: 0,
+      bm: 0,
+    },
+  ],
+  markers: [],
+};
+
+type AnimatedBellProps = Omit<AnimatedIconProps, 'source'>;
+
+export const AnimatedBell: React.FC<AnimatedBellProps> = forwardRef<
```
gewfy, line 352: Can we name this `BellIconAnimated` instead? So we align it with `BellIcon` and `LogoIconAnimated`. Also easier to find them in UI lib since they are sorted alphabetically there :) 
image

gewfy (changes requested):

client/src/lib/components/Buttons/IconButton/AnimatedIconButton.tsx
swemail: true, variant can be set as a state change, nice

client/src/lib/components/Icons/Bell/Bell.tsx
```diff
@@ -12,3 +15,28 @@ export const BellIcon: IconType = ({fill = COLORS.BLACK}) => (
     />
   </Icon>
 );
+
+type BellAnimatedProps = Omit<AnimatedIconProps, 'source'>;
+
+export const BellIconAnimated: React.FC<BellAnimatedProps> = forwardRef<
```
gewfy, line 21: Shouldn't the type be `AnimatedIconType` instead of this custom?

gewfy (approved):
Nice nice nic!

client/src/lib/components/Icons/Bell/Bell.tsx
swemail: Na, that does not type check

PR #2913, opened 2023-05-31
Update edit session modal

transient-jonas:


shared/src/schemas/Session.ts
```diff
@@ -40,6 +40,7 @@ const LiveSessionFieldsSchema = yup
     url: yup.string().required(),
     link: yup.string(),
     inviteCode: yup.number().required(),
+    hostingCode: yup.number(),
```
gewfy, line 43: Before merging this make sure this is not included in any session responses

gewfy (commented):
Know it's WIP but just skimmed through it out of curiosity :) Looks good! 🙌

shared/src/schemas/Session.ts
transient-jonas: Where does the field selection/exclusion happens?
It is taking some time to be able to find it in the code, doesn't seem to be in neither the modal, controller or route 🤔

swemail (approved):
Nice work!

PR #2963, opened 2023-06-08
Dev ergonomics - simplify local star

gewfy:
Removes the need to restart when changing network and keeps the possibility to run on device by starting with `yarn start:device`


swemail (approved):
❤️

PR #2969, opened 2023-06-12
Progress from video

swemail:


gewfy:
If I skip 15sec ahead when reaching the end I seem to move it beyond it's end 
image

gewfy:
The progress indicator goes beyond the track at the end. But since you can't drag it I suggest removing it 
image

gewfy:
Looping videos/lottie animations keep running after the sound has reached it's end.

gewfy:
Progress sometimes doesn't end at 0:00 but instead on a negative number (e.g. -1:57)
image

gewfy:
The video/lottie should be vertically centered
| Design | Implementation |
|--------|--------|
|  |  |

swemail:
> Looping videos/lottie animations keep running after the sound has reached it's end.

I know, but it's another issue I was aiming to fix later. But when is later right, I'll check 👍

gewfy:
I don't think this slide makes any sense also the progress bar for some reason goes beyond it's end. Maybe sync this with @Lisa29k ?
image

gewfy:
For some reason I can't skip past 1:00 on this one 🤷 
image

gewfy (approved):

PR #3127, opened 2023-06-29
Ephemeral user state

gewfy:
Ephemeral user state that can be used before a real user has been set. As soon as a user is created the ephemeral state is migrated into the user.


swemail (approved):
Nice!

PR #3130, opened 2023-06-30
Prevent going to sleep inside session

swemail:


client/src/lib/session/components/Slides/Slides/Blocks/Lottie.tsx
```diff
@@ -91,6 +92,11 @@ const Lottie: React.FC<LottieProps> = ({
     [audioSource],
   );
 
+  useEffect(() => {
+    IdleTimerManager.setIdleTimerDisabled(active);
+    return () => IdleTimerManager.setIdleTimerDisabled(false);
+  }, [active]);
```
gewfy, line 98: Could we make it enabled only when it's actually playing?
image

gewfy (approved):
Boootiful!

PR #3382, opened 2023-08-16
Sticky timer wheel on live sessions

swemail:
Hosts
<image src="https://github.com/29ki/29k/assets/1139207/65fe0651-bb83-434e-a758-ac317ca4e034" width="200" />
Participants
<image src="https://github.com/29ki/29k/assets/1139207/062f788c-afb2-4952-b0c1-d30cbddd98ed" width="200" />


client/src/lib/session/context/TimerContext.ts
```diff
@@ -0,0 +1,10 @@
+import {createContext} from 'react';
+
+export type ProgressTimerContextProps = {
+  onLoad: (duration: number) => void;
+  onSeek: (currentTime: number) => void;
+};
+
+export const ProgressTimerContext = createContext<
+  ProgressTimerContextProps | undefined
+>(undefined);
```
gewfy, line 10: What is the reason to create a context instead of using zustand?

gewfy (approved):
NICE :)

client/src/lib/session/context/TimerContext.ts
swemail: I think cause you happened to mention that as a solution and I wanted to try with props and failed and remembered you mentioning context. I can pretend it's for the ephemeral nature of this state but that is not the real reason :)

PR #3384, opened 2023-08-16
Pic and name from sharing modal

swemail:
https://github.com/29ki/29k/assets/1139207/6267f6a0-839a-4c60-ac93-d618c4072214


gewfy (approved):
Nice nice nice!

PR #3386, opened 2023-08-17
Adjust volume on android

swemail:


client/src/lib/session/hooks/useAdjustVolume.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const useAdjustVolume = () => {
+  return useCallback(async () => {
+    if (Platform.OS === 'android') {
+      const {music, call} = await VolumeManager.getVolume();
+
+      if (call !== music) {
+        VolumeManager.setVolume(Math.max(call ?? 0, music ?? 0), {
+          type: 'music',
+          showUI: false,
```
gewfy, line 13: I was thinking that maybe we should show this, but only when the audio actually starts playing. Not a biggie, maybe for the next iteration...

gewfy (approved):
Added some thoughts, but nothing blocking

client/src/lib/session/hooks/useAdjustVolume.ts
swemail: Show that we adjust it you mean? Yea, I was thinking let's see if this helps with the bug and then refine it a bit

client/src/lib/session/components/Slides/Slides/Content.tsx
```diff
@@ -36,11 +37,19 @@ type ContentProps = {
...
+
+  const content = useMemo(() => {
+    return slide.content ?? {};
+  }, [slide.content]);
+
+  useEffect(() => {
+    if (active && !async && slide.type === 'content') {
+      adjustVolume();
+    }
+  }, [active, async, adjustVolume, slide]);
```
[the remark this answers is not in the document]
swemail: Haha, yea I see, but it is only when we're in a meditation that this should be adjusted. Maybe we could move it out to Session though, and look att current state and if we're actually playing. Then most users have already adjusted volume to hear the host. But don't want to call this more than once and this is a way to assure that

PR #3391, opened 2023-08-21
Global error banner

swemail:
https://github.com/29ki/29k/assets/1139207/58414288-1af8-446c-adc3-d41aadb9e870

https://github.com/29ki/29k/assets/1139207/229ded86-20bc-41ee-9248-eaa292aba032


gewfy:
Nice!! Could you add a UI lib entry so we can see the different variations?

client/src/lib/components/ErrorBanner/ErrorBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const contextValue = useMemo(() => {
+    return {showError: onShowError};
+  }, [onShowError]);
+
+  return (
+    <ErrorBannerContext.Provider value={contextValue}>
+      {children}
+
+      {header && message && (
+        <Container hasAction={Boolean(actionConfig)} top={top}>
```
gewfy, line 103: How about animating this down, so the user sees that animates on top of things. But it should of course just dissapear.

client/src/lib/components/ErrorBanner/ErrorBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const [message, setMessage] = useState('');
+  const [actionConfig, setActionComponent] = useState<ActionConfig | undefined>(
+    undefined,
+  );
+
+  const onShowError = useCallback(
+    (hdr: string, msg: string, action?: ActionConfig) => {
+      setHeader(hdr);
+      setMessage(msg);
+      setActionComponent(action);
```
gewfy, line 76: Should we maybe auto close after X seconds?

gewfy (approved):

PR #3431, opened 2023-08-23
Network and Daily error detection

swemail:
- [x] Show error banner when network is lost (Live and Async)
- [x] Disable slide navigation when network lost (Live)
- [x] Show error banner with rejoin button on daily error (Live)

Testing `useHandleLeaveLiveSession` ans `useNetworkListener` is more work than give value IMO


client/src/lib/session/hooks/useHandleLeaveLiveSession.ts
```diff
@@ -0,0 +1,51 @@
...
+      isReJoining: true,
+    });
+    resetHasFailed();
+  }, [leaveMeeting, resetHasFailed, navigate, session]);
+
+  useEffect(() => {
+    if (hasFailed) {
+      errorBannerContext?.showError(
+        'Session Error',
+        'Pleace rejoin the session',
```
gewfy, line 38: Needs translation and typo in `Pleace`

client/src/lib/navigation/hooks/usePreventGoingBack.ts
```diff
@@ -1,19 +1,23 @@
...
+const usePreventGoingBack = (callback?: () => void, hasFailed?: boolean) => {
   const navigation = useNavigation();
+  const isFocused = useIsFocused();
 
-  useEffect(
-    () =>
-      navigation.addListener('beforeRemove', e => {
+  useEffect(() => {
+    let unsubscribe = () => {};
+    console.log('****************', hasFailed);
```
gewfy, line 10: I know it's a draft, but just so you don't miss it :)

client/src/lib/daily/DailyProvider.tsx
```diff
@@ -81,8 +82,10 @@ const DailyProvider: React.FC<{children: React.ReactNode}> = ({children}) => {
       setParticipantsSortOrder(peerId);
     };
 
-    const onError = ({error}: DailyEventObject<'error'>) => {
-      Sentry.captureException(error);
+    const onError = (errorEvent: DailyEventObject<'error'>) => {
+      //Seems we only get here when it has totally failed
+      setHasFailed();
+      Sentry.captureException(errorEvent.errorMsg);
```
gewfy, line 88: needs a `new Error` too, right?

content/src/ui/Component.NetworkError.json
```diff
@@ -0,0 +1,13 @@
+{
+  "en": {
+    "title": "Network Errors",
+    "message": "App fails to establish or maintain a connection with the server."
```
gewfy, line 4: How about "Lost internet connection" / " It seems like you don't have any connection to the internet at the moment"?

gewfy (commented):
Lajk it!

client/src/lib/session/hooks/useHandleLeaveLiveSession.ts
swemail: That was a draft thingy too :)

client/src/lib/daily/DailyProvider.tsx
swemail: Nice, wasn't there before either so we might have missed these errors

client/src/lib/session/hooks/useUpdateSessionState.ts
```diff
@@ -39,22 +58,32 @@ const useUpdateSessionState = (
...
-        completed,
-      });
+      try {
+        return await sessionApi.updateSessionState(sessionId, {
+          index,
+          playing: false,
+          completed,
+        });
+      } catch (error) {
+        showNetworkRequestError();
```
gewfy, line 68: I was thinking that this should be part of the `apiClient` as a general solution instead of implemented at each place, but with the option to no show errors. E.g. KillSwitch or metrics. WDYS?

client/src/lib/session/hooks/useSubscribeToSession.ts
```diff
@@ -22,16 +28,18 @@ const useSubscribeToSession = (sessionId: LiveSessionType['id']) => {
...
-          console.debug(
-            `Failed to subscribe to live session ${sessionId}`,
-            error,
-          ),
+        error => {
+          errorBannerContext?.showError(
+            t('subscriptionTitle'),
+            t('subscriptionMessage'),
+          );
+          Sentry.captureException(error, {extra: {sessionId}});
```
gewfy, line 36: Nice that we start capturing these

PR #3464, opened 2023-08-24
Kick user out

swemail:
<image src="https://github.com/29ki/29k/assets/1139207/5bb89ddc-f85b-44a4-b925-8599e3fe9ce7" width="200" />

https://github.com/29ki/29k/assets/1139207/242c50c7-53e3-43a7-adef-e1cf0a6bbd13


functions/src/api/sessions/index.ts
```diff
@@ -216,6 +217,31 @@ sessionsRouter.put(
   },
 );
 
+sessionsRouter.put(
+  '/removeMyself',
```
gewfy, line 221: Shouldn't this be `:sessionId/removeMyself`?

gewfy (changes requested):

gewfy (approved):
Nice!

PR #3578, opened 2023-09-01
Exercise feedback count

transient-jonas:

### DIFF
--- a/client/src/lib/sessions/api/session.ts
+++ b/client/src/lib/sessions/api/session.ts
@@ -1,3 +1,4 @@
+import {isNil, reject} from 'ramda';
 import {
   LiveSessionType,
   SessionMode,
@@ -225,8 +226,9 @@ export const getFeedbackCountByExercise = async (
   exerciseId: string,
   mode: SessionMode,
 ): Promise<{positive: number; negative: number}> => {
+  const queryParams = new URLSearchParams(reject(isNil, {mode}));
   const response = await apiClient(
-    `${SESSIONS_ENDPOINT}/exercise/${exerciseId}/${mode}/feedback/count`,
+    `${SESSIONS_ENDPOINT}/exercises/${exerciseId}/rating?${queryParams}`,
     {
       method: 'GET',
     },
--- a/client/src/lib/sessions/api/session.ts
+++ b/client/src/lib/sessions/api/session.ts
@@ -1,5 +1,6 @@
 import {
   LiveSessionType,
+  SessionMode,
   SessionStateType,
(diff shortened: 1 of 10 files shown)


client/src/lib/sessions/api/session.ts
```diff
@@ -219,3 +220,21 @@ export const acceptHostingInvite = async (
 
   return response.json();
 };
+
+export const getFeedbackCountByExercise = async (
+  exerciseId: string,
+  mode: SessionMode,
+): Promise<{positive: number; negative: number}> => {
+  const response = await apiClient(
+    `${SESSIONS_ENDPOINT}/exercise/${exerciseId}/${mode}/feedback/count`,
```
gewfy, line 229: I beleive this should be plural `exercises` as with `sessions`. May I suggest this instead:
`sessions/exercises/${exerciseId}/rating?mode=${mode}`?

The reason to put `mode`as a parameter is that it's param (like `limit`) rather than a resource and will probably be optional in the future.

functions/src/models/metrics.ts
```diff
@@ -69,3 +71,19 @@ export const addFeedback = async (feedback: Feedback) => {
...
+  exerciseId: string,
+  mode?: SessionMode,
+) => {
+  let query = firestore()
+    .collection(FEEDBACK_COLLECTION)
+    .where('exerciseId', '==', exerciseId);
+
+  if (mode) {
+    query = query.where('mode', '==', mode);
+  }
```
gewfy, line 85: Do we have an index for this combo?

client/src/routes/overlays/SessionOverlay/SessionOverlay.tsx
```diff
@@ -239,6 +257,12 @@ const SessionOverlay = () => {
         <Spacer16 />
 
         <Content>
+          {count && count.positive > 0 && (
+            <CenteredRow>
+              <FeedbackThumb />
+              <FeedbackCount>{count.positive}</FeedbackCount>
+            </CenteredRow>
+          )}
```
gewfy, line 265: Could you add screenshots on this? :)

gewfy (changes requested):

functions/src/models/metrics.ts
transient-jonas: Good reminder! haven't added to the indexes yet! will do it now 👍

gewfy (approved):
Nice! Understand that the design doesn't match the new overlay look, but how about moving the thumb up to float center aligned (vertically) with the close button? It pushes the content down a lot right now.

PR #3686, opened 2023-09-13
Related exercises list

transient-jonas:
https://github.com/29ki/29k/assets/7523828/aa2c1ab2-0868-4574-b592-55fabc5c1f31


client/src/lib/content/hooks/useGetExercisesByTags.tsx
```diff
@@ -0,0 +1,28 @@
...
+      .filter(e => showHiddenContent || !e.hidden)
+      .filter(
+        e =>
+          intersection(
+            (e.tags || []).map(t => t.tag),
+            tagNames,
+          ).length > 0 && e.id !== excludeId,
+      );
+
+    return take(5, exerciseList);
```
gewfy, line 24: Can we remove the limiting from here and do it outside of the hook? Or make it an argument. We will be using this hook when filtering all exercises in explore soon.

gewfy: Another approach could be to split this into one pure `useGetExercisesByTags` that doesn't exclude or limit. And then a `useGetRelatedExercises` that uses `useGetExercisesByTags` but also excludes and limits the result.

client/src/lib/content/hooks/useGetExercisesByTags.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const useGetExercisesByTags = (tags: Tag[], excludeId?: string) => {
+  const {showHiddenContent} = useAppState(state => state.settings);
+  const exercises = useExercises();
+  const tagNames = useMemo(() => tags?.map(t => t.tag), [tags]);
+
+  return useMemo(() => {
+    if (!tagNames) return [];
+    const exerciseList = exercises
+      .filter(e => showHiddenContent || !e.hidden)
```
gewfy, line 15: This filter is unnecessary because it's handled in i18n. If you see hidden content even though the toggle is off, try restarting the app.

gewfy (changes requested):

gewfy (approved):

PR #3732, opened 2023-09-20
Add feedback params and improve types

transient-jonas:


shared/src/schemas/Feedback.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const FeedbackSchema = FeedbackInputSchema.concat(
+  yup.object().shape({
+    id: yup.string().required(),
+    approved: yup.boolean(),
+    createdAt: transformTimestamp.required(),
+  }),
+);
+
+export const CreateFeedbackBodySchema = FeedbackInputSchema.concat(
```
gewfy, line 45: Can we rename this to `AddFeedbackBodySchema`?

shared/src/types/Feedback.ts
```diff
@@ -1,21 +1,13 @@
...
+import {
+  CreateFeedbackBodySchema,
+  FeedbackParamsSchema,
+  FeedbackInputSchema,
+  FeedbackSchema,
+} from '../schemas/Feedback';
+
+export type FeedbackParams = yup.InferType<typeof FeedbackParamsSchema>;
+export type FeedbackInput = yup.InferType<typeof FeedbackInputSchema>;
+export type CreateFeedbackBody = yup.InferType<typeof CreateFeedbackBodySchema>;
```
gewfy, line 12: `AddFeedbackBody`

functions/src/metrics/logFeedback/index.ts
```diff
@@ -1,40 +1,18 @@
...
-
 export const feedbackRouter = router.post(
   '/',
   validator({
-    body: feedbackBodySchema,
+    body: CreateFeedbackBodySchema,
   }),
   async ({request, response}) => {
-    const feedback = request.body as Body;
+    const feedback = request.body as CreateFeedbackBody;
```
gewfy, line 15: Destruct `body` instead of `request.body` and it should give you the correct type instead :)

functions/src/models/slack/messageBlocks.ts
```diff
@@ -246,7 +248,11 @@ export const createFeedbackBlocks = (
         `*Exercise:*\n${exercise} (${sessionMode ? sessionMode : ''} - ${
           sessionType ? sessionType : ''
         })\n\n` +
-        `*Comment:*\n${comment}`,
+        `*Comment:*\n${comment}\n\n` +
+        `*Model:* ${params.model ?? ''}\n\n` +
+        `*OS:* ${params.os ?? ''}\n\n` +
+        `*Native Version:* ${params.nativeVersion ?? ''}\n\n` +
+        `*Bundle Version:* ${params.bundleVersion ?? ''}`,
```
gewfy, line 255: This will probably swell the message vertically a lot. But let's keep an eye on it and maybe put it in two columns if needed. There's a specific slack message block for that. Like with the old app
image

shared/src/schemas/Feedback.ts
```diff
@@ -0,0 +1,47 @@
+import * as yup from 'yup';
+
+import {SessionMode, SessionType} from './Session';
+import {transformTimestamp} from '../modelUtils/transform';
+
+export const FeedbackParamsSchema = yup.object().shape({
+  screen: yup.string(),
+  key: yup.string(),
```
gewfy, line 8: I can't see screen or key being sent from the client? But also I don't think it makes sense as we ask for feedback at the same places every time. So I believe it's safe to remove completely :)

gewfy (changes requested):

functions/src/metrics/logFeedback/index.ts
transient-jonas: Screenshot 2023-09-21 at 15 23 56

transient-jonas: seems like we use `request.body as ...`  across the codebase?
Screenshot 2023-09-21 at 15 25 04

gewfy: Something off with the validation lib... resolving for now

gewfy (approved):

PR #3736, opened 2023-09-25
Add show more text component

transient-jonas:
https://github.com/29ki/29k/assets/7523828/e4a56767-acbd-47a3-9719-73e495a83432

### DIFF
--- a/client/src/lib/components/ShowMoreText/ShowMoreText.tsx
+++ b/client/src/lib/components/ShowMoreText/ShowMoreText.tsx
@@ -1,0 +1,33 @@
+import React from 'react';
+
+import ReadMore from '@fawazahmed/react-native-read-more';
+
+import {COLORS} from '../../../../../shared/src/constants/colors';
+
+import {useTranslation} from 'react-i18next';
+
+import {Body16} from '../Typography/Body/Body';
+
+const readMoreTextStyles = {color: COLORS.PRIMARY};
+
+const ShowMoreText: React.FC<{
+  numberOfLines?: number;
+  children: React.ReactNode;
+}> = ({numberOfLines = 5, children}) => {
+  const {t} = useTranslation('Component.ShowMoreText');
+
+  return (
+    <ReadMore
+      customTextComponent={Body16}
+      numberOfLines={numberOfLines}
+      seeLessText={t('shrinkLabel')}
+      seeMoreText={t('expandLabel')}
+      seeMoreStyle={readMoreTextStyles}
+      seeLessStyle={readMoreTextStyles}
+      animate={false}>
+      {children}
+    </ReadMore>
+  );
+};
+
+export default ShowMoreText;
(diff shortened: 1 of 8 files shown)


client/src/lib/components/ShowMoreText/ShowMoreText.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+import ReadMore from '@fawazahmed/react-native-read-more';
```
gewfy, line 3: Could you explain the benefits of this lib? Couldn't we just do this with a character count and simple state in this component? The lib is a bit sloppy built (e.g. hard to find the source code) so I'm a bit hesitant at adding it.

transient-jonas: aha, I approached it similarly (mindset wise) using
```typescript
onTextLayout={({ nativeEvent: { lines } }) =>
    setShowMoreButton(lines.length > numberOfLines)
    ...
```

But it didn't work well also couldn't add the show more button by the end of the last line, would be interesting trying with character count as you mentioned

transient-jonas: on the example above I'm using `numberOfLines={5}` on the text to limit

transient-jonas: Expo seems to have a lib for this as well https://github.com/expo/react-native-read-more-text

transient-jonas: They seem to use an api to find out the height of the text component
https://github.com/expo/react-native-read-more-text/blob/3e63ca30bd0b7925893cf76bc015ab7e9b7a75a3/index.js#L103-L105

transient-jonas: Shall I copy their approach and do a component in our codebase? wdyt? @gewfy

gewfy: Thanks for doing some research here! I believe we did something similar, height based, in the old. Let me check :)

gewfy: Yeach exactly, in the old ap we did what you proposed ⬆️ 
```js
                onTextLayout={({ nativeEvent: { lines } }) => {
                  if (lines.length > NUMBER_OF_LINES && isNil(truncated)) {
                    setTruncated(true);
                  }
                }}
```

And then just absolute position a `Show more` at the right bottom of that component, or bellow.

gewfy: Since this seems more complex than it looks on the surface I leave it up to you. Fine to go with the lib you chose too :)

client/src/routes/modals/CompletedSessionModal/CompletedSessionModal.tsx
```diff
@@ -259,7 +259,7 @@ const CompletedSessionModal = () => {
           <>
             <Spacer16 />
             <Gutters>
-              <Markdown>{exercise?.description}</Markdown>
+              <ShowMoreText>{exercise?.description}</ShowMoreText>
```
gewfy, line 262: One tiny thing. Aren't we loosing the possibility to use markdown with this? Wrap it in `<ShowMoreText>`?

gewfy (approved):

client/src/routes/modals/CompletedSessionModal/CompletedSessionModal.tsx
transient-jonas: Markdown doesn't seem to support `numberOfLines` or similar functionality, does it? 🤔

transient-jonas: ah you mean using the library, I need to try that out

PR #4424, opened 2023-12-12
Migrate to decap-cms

gewfy:
Ping @Greenheart, have you tried?


Greenheart:
Thanks @gewfy! I've experimented with it several times but got stuck with bugs in the new version so never ended up finishing the migration. I have considered [Keystatic](https://github.com/Thinkmill/keystatic) and [sveltiacms](https://github.com/sveltia/sveltia-cms) for other projects and they are both quite promising.

gewfy:
I haven't found any bugs so far though 🤔

PR #5733, opened 2024-08-18
Early web version

gewfy:
Simple early web version of the content


Greenheart:
@gewfy Finally! :open_mouth: 

Even though I'm no longer involved actively with the IDGs, this would be lovely to add to the idg.tools website to give more traffic to all the useful exercises and tools that have been created for Aware!

PR #6362, opened 2024-12-06
Temporary fix for modal being on top of some action menus

gewfy:
... until Bottom Sheet FullWindowOverlay issue is reported and fixed


gewfy:
Reported here https://github.com/th3rdwave/react-navigation-bottom-sheet/issues/31

